PR #33, opened 2017-07-17
Implement save files and json serialization

nightpool:
Things that don't currently work:

 - [x] saving a color sometimes is a little wonky because it tries to serialize the internalPaint attribute, this can be fixed with a custom ColorSaver that should be easy to write
 - [ ] no testfx tests for saving and loading--I tried to get these to work, but I couldn't figure out a way to script the file loader that worked cross platform. Maybe somebody else could take a look at this?

### DIFF
--- a/src/main/java/edu/wpi/first/shuffleboard/json/WidgetSaver.java
+++ b/src/main/java/edu/wpi/first/shuffleboard/json/WidgetSaver.java
@@ -1,0 +1,42 @@
+package edu.wpi.first.shuffleboard.json;
+
+import com.google.gson.JsonDeserializationContext;
+import com.google.gson.JsonElement;
+import com.google.gson.JsonObject;
+import com.google.gson.JsonParseException;
+import com.google.gson.JsonSerializationContext;
+import edu.wpi.first.shuffleboard.sources.Sources;
+import edu.wpi.first.shuffleboard.widget.Widget;
+import edu.wpi.first.shuffleboard.widget.Widgets;
+import javafx.beans.property.Property;
+
+@AnnotatedTypeAdapter(forType = Widget.class)
+public class WidgetSaver implements ElementTypeAdapter<Widget> {
+  @Override
+  public JsonElement serialize(Widget src, JsonSerializationContext context) {
+    JsonObject object = new JsonObject();
+    object.addProperty("_type", src.getName());
+    object.addProperty("_source", src.getSource().getId());
+    for (Property p : src.getProperties()) {
+      object.add(p.getName(), context.serialize(p.getValue()));
+    }
+    return object;
+  }
+
+  @Override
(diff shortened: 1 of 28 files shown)


src/main/java/edu/wpi/first/shuffleboard/sources/SourceType.java
```diff
@@ -35,4 +45,16 @@ public String removeProtocol(String text) {
     }
   }
 
+  public static DataSource<?> fromUri(String uri) {
```
SamCarlberg: This should be in the `Sources` class

nightpool:
Disagree. "Sources" conceptually represents the existing sources that have
been created in some way or another. fromUri is a method that doesn't worry
about that, it just deals with dispatches to different source types. One is
at a lower level of abstraction then the other.

SamCarlberg:
This would be a better solution, then

```java
public DataSource<?> forUri(String uri) {
  if (!uri.startsWith(protocol)) {
    throw new IllegalArgumentException("URI does not start with the correct protocol: " + uri);
  }
  return sourceSupplier.apply(removeProtocol(uri));
}

public static DataSource<?> fromUri(String uri) {
  return Stream.of(values())
    .filter(type -> type != NONE)
    .filter(type -> uri.startsWith(type.protocol))
    .findFirst()
    .map(type -> type.forUri(uri))
    .orElseGet(EmptyDataSource::new);
}

```

nightpool:
will change

src/main/java/edu/wpi/first/shuffleboard/json/WidgetSaver.java
```diff
@@ -0,0 +1,42 @@
...
+import edu.wpi.first.shuffleboard.widget.Widget;
+import edu.wpi.first.shuffleboard.widget.Widgets;
+import javafx.beans.property.Property;
+
+@AnnotatedTypeAdapter(forType = Widget.class)
+public class WidgetSaver implements ElementTypeAdapter<Widget> {
+  @Override
+  public JsonElement serialize(Widget src, JsonSerializationContext context) {
+    JsonObject object = new JsonObject();
+    object.addProperty("_type", src.getName());
```
nightpool: So as to avoid conflicts with other properties of the same name?

SamCarlberg: The extra properties could go in a nested map. Or we just blacklist "type" and "source" as property names

nightpool: I don't think the extra complexity is worth it for two properties.

nightpool:
fixes pushed.

SamCarlberg:
Failing tests are noted in #37

PR #34, opened 2017-07-17
Fix network table tree bugs

### LINKED ISSUE #31: Rapidly changing network table entries cannot be dragged as sources
Instead, a different entry is chosen when dragging so the resulting source/widget is not the one the user selected in the tree

SamCarlberg:
- Ordering is now done ignoring capitalization
- Fix dragging incorrect sources when the tree is updating

Fixes #31


AustinShalit:
@SamCarlberg Do we need to make these changes to OutlineViewer too?

SamCarlberg:
@AustinShalit Currently no. It's only an issue if you want to drag a row of the tree

PR #36, opened 2017-07-19
Set initial window size to screen's visual bounds

calcmogul:
This prevents my computer's task bar from covering the window decorations and
tab list.

### DIFF
--- a/src/main/java/edu/wpi/first/shuffleboard/Shuffleboard.java
+++ b/src/main/java/edu/wpi/first/shuffleboard/Shuffleboard.java
@@ -8,9 +8,11 @@ import java.io.IOException;
 
 import javafx.application.Application;
 import javafx.fxml.FXMLLoader;
+import javafx.geometry.Rectangle2D;
 import javafx.scene.Scene;
 import javafx.scene.layout.Pane;
 import javafx.scene.text.Font;
+import javafx.stage.Screen;
 import javafx.stage.Stage;
 
 @SuppressWarnings("JavadocMethod")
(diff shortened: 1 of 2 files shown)


SamCarlberg:
This would break for people with multiple screens with different resolutions, like plugging in a macbook to an iMac as a secondary display.

I wouldn't be against binding the maximum width and height to the size of the current display, but that would run into issues when dragging the window between two screens with different resolutions

calcmogul:
Perhaps we could give the window a sane default size, like the size of the current display, and not set a minimum or maximum.

AustinShalit:
@calcmogul - How does the window get that large? Does it start that way?

calcmogul:
Yes. Since the size specified in the XML is infinite, the window maximizes the window area to my whole screen, which also hides the window decorations. My taskbar then covers the tab list. I use the Xfce desktop environment.

SamCarlberg:
The _maximum_ size is infinite, but the default size should be based on the size of the components in the window.

Are you using a high DPI display? I know JavaFX has issues with DPI scaling on Linux.

calcmogul:
For some reason, shuffleboard starts at the maximum size. I've changed `maxWidth` and `maxHeight` in MainWindow.fxml to verify this is the case. My display is 1366x768, so definitely not high DPI.

nightpool:
We definitely don't want to be overriding the WMs standard behavior here. I think #55 fixes the pain point here in a less obtrusive way (see e01ea08)

AustinShalit:
Does #55 fix the issue this PR was trying to solve @calcmogul?

calcmogul:
No. The width is sane now, but the window is still too tall.

Issue #49, opened 2017-07-19
Launches sized too large


Kevin-OConnor:
Running shuffleboard on my machine results in it launching with the window size too large. It flows off the top and bottom of the screen, hiding the menus.

SamCarlberg:
Operating system?

Kevin-OConnor:
Windows 7 x64. I also just noticed a few errors in the Gradle window, not sure if they are related or not:
Exception in thread "JavaFX-Launcher" java.nio.file.InvalidPathException: Illegal char <:> at index 66: C:\Users\koconnor/SmartDashboard/dashboard_recording_2017-07-19_13:38:25.frc
        at sun.nio.fs.WindowsPathParser.normalize(WindowsPathParser.java:182)
        at sun.nio.fs.WindowsPathParser.parse(WindowsPathParser.java:153)
        at sun.nio.fs.WindowsPathParser.parse(WindowsPathParser.java:77)
        at sun.nio.fs.WindowsPath.parse(WindowsPath.java:94)
        at sun.nio.fs.WindowsFileSystem.getPath(WindowsFileSystem.java:255)
        at java.nio.file.Paths.get(Paths.java:84)
        at edu.wpi.first.shuffleboard.sources.recording.Serialization.saveRecording(Serialization.java:107)
        at edu.wpi.first.shuffleboard.sources.recording.Recorder.saveToDisk(Recorder.java:60)
        at edu.wpi.first.shuffleboard.sources.recording.Recorder.lambda$new$98(Recorder.java:41)
        at edu.wpi.first.shuffleboard.sources.recording.Recorder$$Lambda$82/18330560.changed(Unknown Source)
        at com.sun.javafx.binding.ExpressionHelper$SingleChange.fireValueChangedEvent(ExpressionHelper.java:182)
        at com.sun.javafx.binding.ExpressionHelper.fireValueChangedEvent(ExpressionHelper.java:81)
        at javafx.beans.property.BooleanPropertyBase.fireValueChangedEvent(BooleanPropertyBase.java:103)
        at javafx.beans.property.BooleanPropertyBase.markInvalid(BooleanPropertyBase.java:110)
        at javafx.beans.property.BooleanPropertyBase.set(BooleanPropertyBase.java:144)
        at edu.wpi.first.shuffleboard.sources.recording.Recorder.setRunning(Recorder.java:146)
        at edu.wpi.first.shuffleboard.sources.recording.Recorder.start(Recorder.java:94)
[...]

PR #55, opened 2017-07-19
Dynamically add and remove rows and columns from tile panes as they're resized

### LINKED ISSUE #7: The tile grid should add or remove rows and columns as it's resized
Currently the amounts of rows and columns are hardcoded. Columns should be added or removed as the tile grid is resized, possibly down to some fixed minimum size or to a size where anything smaller would clip or remove tiles

SamCarlberg:
Fixes #7 and #49


src/main/java/edu/wpi/first/shuffleboard/components/TilePane.java
```diff
@@ -29,7 +29,8 @@
 
   private static final int DEFAULT_COL_COUNT = 1;
   private static final int DEFAULT_ROW_COUNT = 1;
-  private static final double MIN_TILE_SIZE = 32; // pixels
+  private static final double MIN_TILE_SIZE = 64; // pixels
+  private static final double MAX_TILE_SIZE = 65535; // pixels
```
nightpool: reasoning?

SamCarlberg: For which?

nightpool: such a large max tile size. When would this ever happen? Why is this a good thing to have in the code? this is larger then any screen I know of.

SamCarlberg: Because the constraints need a maximum size and the default is Infinity

nightpool: the default is *negative* infinity, which we should use with Region.USE_PREF_SIZE rather then creating a new constant here.

src/main/java/edu/wpi/first/shuffleboard/components/TilePane.java
```diff
@@ -29,7 +31,7 @@
 
   private static final int DEFAULT_COL_COUNT = 1;
   private static final int DEFAULT_ROW_COUNT = 1;
-  private static final double MIN_TILE_SIZE = 32; // pixels
+  private static final double MIN_TILE_SIZE = 64; // pixels
```
JLLeitschuh: Number of pixels is a bad size metric as it can vary wildly. Shouldn't you use `em` instead?
https://stackoverflow.com/a/609561/3708426

SamCarlberg: JavaFX doesn't have a notion of `%` or `em` sizing. Saying this is pixels might be a misnomer because of the scaling that JavaFX does automatically for high DPI displays

JLLeitschuh: Right, okay, that makes sense.

SamCarlberg:
Made widgets move and resize when needed. May want to stick the widget panes in scrollpanes as well, but I think if you have widgets off the screen they should be in separate tabs instead, especially since #57 makes working with tabs so much easier.

There's still a bug when multiple rows/columns get removed at once that a widget will get stranded outside the grid. Not too hard to solve though.

SamCarlberg:
@nightpool Can you please re-review?

nightpool:
Looks like we don't correctly check **all** dependents if we find a tile that can collapse:

collapse

src/main/java/edu/wpi/first/shuffleboard/WidgetPaneController.java
```diff
@@ -95,6 +109,54 @@ private void initialize() {
...
+        return;
+      }
+      if (newCount < oldCount) {
+        // shift and shrink tiles to the left
+        pane.getTiles().stream()
+            .filter(tile -> {
+              final TileLayout layout = pane.getTileLayout(tile);
+              return layout.origin.col + layout.size.getWidth() > newCount;
+            })
+            .forEach(tile -> moveTileLeft(tile, oldCount - newCount));
```
nightpool: This will move the first tile left N times, then the second tile, etc. I think it's probably better to move EVERY tile once, then every tile again, etc. Basically, maybe move the `for (int i = 0; i < count; i++)` to this function instead of the moveTileLeft.

src/main/java/edu/wpi/first/shuffleboard/WidgetPaneController.java
```diff
@@ -241,4 +303,104 @@ private Menu createChangeMenus(WidgetTile tile) {
...
+        break;
+      }
+    }
+  }
+
+  private Optional<Runnable> moveTileLeft(WidgetTile tile) {
+    TileLayout layout = pane.getTileLayout(tile);
+    final int previousColumn = layout.origin.col - 1;
+    if (!pane.isOverlapping(previousColumn, layout.origin.row,
+        layout.size.getWidth(), layout.size.getHeight(), n -> tile == n) && previousColumn >= 0) {
```
nightpool: can we make this just layout.size? and can we use previousPosition instead of previousColumn, layout.origin.row?

src/main/java/edu/wpi/first/shuffleboard/WidgetPaneController.java
```diff
@@ -241,4 +303,104 @@ private Menu createChangeMenus(WidgetTile tile) {
...
+    } else if (layout.origin.col > 0) {
+      // Check to see if we can move any tiles to the left, if possible
+      List<Runnable> runs = IntStream.range(layout.origin.row, layout.origin.row + layout.size.getHeight())
+          .mapToObj(row -> pane.tileAt(previousColumn, row))
+          .filter(Optional::isPresent) // guaranteed to be at least one tile
+          .map(Optional::get)
+          .distinct() // need to make sure we have no repeats, or n-row tiles will get moved n times
+          .filter(t -> tile != t)
+          .map(this::moveTileLeft) // recursion here
+          .filter(Optional::isPresent)
```
nightpool: filter is the cause of the bug here. We need to ensure that ALL of the tiles to the left can move, not just ONE of them.

src/main/java/edu/wpi/first/shuffleboard/WidgetPaneController.java
```diff
@@ -241,4 +303,104 @@ private Menu createChangeMenus(WidgetTile tile) {
...
+      if (move.isPresent()) {
+        move.get().run();
+      } else {
+        // Can't move any further, bail
+        break;
+      }
+    }
+  }
+
+  private Optional<Runnable> moveTileUp(WidgetTile tile) {
```
nightpool: lots of duplication here between moveTileLeft and moveTileUp. this take a direction to move in, which we can then subtract from the current tile

src/main/java/edu/wpi/first/shuffleboard/WidgetPaneController.java
```diff
@@ -241,4 +303,104 @@ private Menu createChangeMenus(WidgetTile tile) {
...
+          .filter(Optional::isPresent) // guaranteed to be at least one tile
+          .map(Optional::get)
+          .distinct() // need to make sure we have no repeats, or n-row tiles will get moved n times
+          .filter(t -> tile != t)
+          .map(this::moveTileUp) // recursion here
+          .filter(Optional::isPresent)
+          .map(Optional::get)
+          .collect(Collectors.toList());
+      if (runs.isEmpty()) {
+        // Can't move or shrink any tiles to the left of this one to make it move
```
nightpool: The preference here is weird. We should probably try shrinking the **current** tile before we shrink ones to the left of it, because that seems more intuitive.

src/main/java/edu/wpi/first/shuffleboard/WidgetPaneController.java
```diff
@@ -241,4 +303,104 @@ private Menu createChangeMenus(WidgetTile tile) {
...
+          .filter(Optional::isPresent) // guaranteed to be at least one tile
+          .map(Optional::get)
+          .distinct() // need to make sure we have no repeats, or n-row tiles will get moved n times
+          .filter(t -> tile != t)
+          .map(this::moveTileLeft) // recursion here
+          .filter(Optional::isPresent)
+          .map(Optional::get)
+          .collect(Collectors.toList());
+      if (runs.isEmpty()) {
+        // Can't move or shrink any tiles to the left of this one to make it move
```
nightpool: The preference here is weird. We should probably try shrinking the **current** tile before we shrink ones to the left of it, because that seems more intuitive.

src/main/java/edu/wpi/first/shuffleboard/components/TilePane.java
```diff
@@ -332,38 +355,49 @@ public boolean isOpen(GridPoint point, TileSize tileSize, Predicate<Node> ignore
...
+   * ignoring some nodes when calculating collisions. Note that this method does <i>not</i> perform bounds checking;
+   * use {@link #isOpen(int, int, int, int, Predicate) isOpen} to check if a widget can be placed at that point.
+   *
+   * @param col        the column index of the point to check
+   * @param row        the row index of the point to check
+   * @param tileWidth  the width of the tile
+   * @param tileHeight the height of the tile
+   * @param ignore     a predicate to use to ignore nodes when calculating collisions
+   */
+  public boolean isOverlapping(int col, int row, int tileWidth, int tileHeight, Predicate<Node> ignore) {
```
nightpool: should take a TileLayout instead of 4 ints.

src/main/java/edu/wpi/first/shuffleboard/util/RoundingMode.java
```diff
@@ -0,0 +1,27 @@
...
+  RoundingMode DOWN = value -> (int) Math.floor(value);
+
+  /**
+   * Rounds a fractional value to an integer.
+   *
+   * @param value the value to round
+   */
+  int round(double value);
+
+
```
nightpool: extra newline

nightpool:
Has one conflict, should be rebased against master so we don't get codecov commits when we run the tests.

calcmogul:
The latest version of this PR fixed #49 for me.

src/main/java/edu/wpi/first/shuffleboard/Shuffleboard.java
```diff
@@ -40,6 +41,10 @@ public void init() throws Exception {
   @Override
   public void start(Stage primaryStage) throws IOException {
     primaryStage.setScene(new Scene(mainPane));
+    primaryStage.setMinWidth(640);
+    primaryStage.setMinHeight(480);
+    primaryStage.setWidth(Screen.getPrimary().getBounds().getWidth());
+    primaryStage.setHeight(Screen.getPrimary().getBounds().getHeight());
```
nightpool: should be getVisualBounds per https://github.com/wpilibsuite/shuffleboard/pull/55#issuecomment-318926171

SamCarlberg: Already done. Working on bug fixes atm, will push when that's done

src/main/java/edu/wpi/first/shuffleboard/WidgetPaneController.java
```diff
@@ -257,4 +316,88 @@ private Menu createChangeMenus(WidgetTile tile) {
...
+          .distinct() // need to make sure we have no repeats, or n-row tiles will get moved n times
+          .filter(t -> tile != t)
+          .map(this::moveTileLeft) // recursion here
+          .collect(Collectors.toList());
+      if (runs.stream().allMatch(Optional::isPresent)) {
+        return Optional.of(() -> {
+          runs.forEach(o -> o.get().run());
+          GridPane.setColumnIndex(tile, previousColumn);
+        });
+      } else {
```
nightpool: This still has the shrinking logic in the wrong order. It shrinks things from left-to-right, which is very confusing—you're moving the window from right-to-left. It should shrink the most immediate thing first, and *then* try and recurse.

PR #57, opened 2017-07-21
Add menu items and shortcuts to add, remove, and navigate to tabs

SamCarlberg:
@nightpool or @AustinShalit can you test this on macOS?

### DIFF
--- a/src/main/java/edu/wpi/first/shuffleboard/MainWindowController.java
+++ b/src/main/java/edu/wpi/first/shuffleboard/MainWindowController.java
@@ -1,6 +1,8 @@
 package edu.wpi.first.shuffleboard;
 
 import com.google.common.io.Files;
+
+import edu.wpi.first.shuffleboard.components.AdderTab;
 import edu.wpi.first.shuffleboard.components.DashboardTabPane;
 import edu.wpi.first.shuffleboard.components.WidgetGallery;
 import edu.wpi.first.shuffleboard.dnd.DataFormats;
@@ -18,6 +20,7 @@ import edu.wpi.first.shuffleboard.util.Storage;
 import edu.wpi.first.shuffleboard.widget.NetworkTableTreeWidget;
 import edu.wpi.first.shuffleboard.widget.Widget;
 import edu.wpi.first.shuffleboard.widget.Widgets;
+
 import javafx.beans.value.ObservableValue;
 import javafx.fxml.FXML;
 import javafx.geometry.Insets;
@@ -125,6 +128,16 @@ public class MainWindowController {
     sourcesTab.setContent(networkTables.getView());
 
     widgetGallery.loadWidgets(Widgets.allWidgets());
+
+    root.addEventFilter(KeyEvent.KEY_PRESSED, e -> {
+      if (e.isShortcutDown() && e.getCode().isDigitKey()) {
+        // DIGIT1 is named "1", DIGIT2 is "2", etc.
+        int tabIndex = Integer.valueOf(e.getCode().getName()) - 1;
+        if (dashboard.getTabs().size() >= tabIndex) {
+          dashboard.getSelectionModel().select(tabIndex);
+        }
+      }
+    });
   }
 
   private void makeSourceRowDraggable(TreeTableRow<? extends SourceEntry> row) {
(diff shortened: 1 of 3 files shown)


AustinShalit:
Trying to select an invalid tab results in selecting the first tab.  Other than that, works fine for me.

AustinShalit:
Oh, if there is only one tab, the application should close if I use the close shortcut.

src/main/java/edu/wpi/first/shuffleboard/MainWindowController.java
```diff
@@ -125,6 +128,16 @@ private void initialize() throws IOException {
     sourcesTab.setContent(networkTables.getView());
 
     widgetGallery.loadWidgets(Widgets.allWidgets());
+
+    root.addEventFilter(KeyEvent.KEY_PRESSED, e -> {
+      if (e.isShortcutDown() && e.getCode().isDigitKey()) {
+        // DIGIT1 is named "1", DIGIT2 is "2", etc.
+        int tabIndex = Integer.valueOf(e.getCode().getName()) - 1;
+        if (dashboard.getTabs().size() >= tabIndex) {
+          dashboard.getSelectionModel().select(tabIndex);
```
nightpool: this would probably be more correct to encapsulate inside of a void selectDashboard(int n) function, because there are presentational tabs we don't want to select (the adder tab)

nightpool:
@AustinShalit disagree. The correct behavior should be to close the last tab and create a new, blank tab. Is that the current behavior for you?

src/main/java/edu/wpi/first/shuffleboard/MainWindowController.java
SamCarlberg: That would work. Would be nice if we want something like `ALT+LEFT` and `ALT+RIGHT` to move between consecutive tabs

AustinShalit:
That is the current behavior for me.  I disagree with that behavior. We should not disrupt the operating system's way of doing things.  On a Mac, I expect the main window to close when I close the last tab.  This is true for every application that supports [Tabbed Windows](https://developer.apple.com/documentation/appkit/windows_panels_and_screens/tabbed_windows).  On Windows I expect the same behavior.  Even the FRC Driver Station closes when sending it the keyboard shortcut to close the window.

The only apps where I do not expect this behavior is IDEs like Eclipse and InteliJ.

src/main/java/edu/wpi/first/shuffleboard/components/DashboardTabPane.java
```diff
@@ -60,8 +66,8 @@ public void addWidgetToActivePane(Widget widget) {
    */
   public void selectWidgets(Predicate<Widget> selector) {
     getTabs().stream().map(optionalCast(DashboardTab.class))
-            .forEach(tab -> tab.map(DashboardTab::getWidgetPane)
-                    .ifPresent(pane -> pane.selectWidgets(selector)));
+        .forEach(tab -> tab.map(DashboardTab::getWidgetPane)
+            .ifPresent(pane -> pane.selectWidgets(selector)));
```
JLLeitschuh: You should consider at least using an editor config file to make this consistent:
http://editorconfig.org/

src/main/java/edu/wpi/first/shuffleboard/MainWindowController.java
```diff
@@ -320,4 +333,19 @@ private void loadPlayback() throws IOException {
     Playback playback = Playback.load(selected.getAbsolutePath());
     playback.start();
   }
+
+  @FXML
+  private void closeCurrentTab() {
+    Tab current = dashboard.getSelectionModel().getSelectedItem();
+    if (!(current instanceof AdderTab)) {
```
nightpool: should be current instanceof DashboardTab

SamCarlberg: makes sense

src/main/java/edu/wpi/first/shuffleboard/MainWindowController.java
```diff
@@ -131,11 +131,15 @@ private void initialize() throws IOException {
...
-          dashboard.getSelectionModel().select(tabIndex);
+        int tabIndex;
+        if (e.getCode().getName().contains("Numpad")) {
+          // remove leading "Numpad " to parse the number
+          tabIndex = Integer.valueOf(e.getCode().getName().substring(7)) - 1;
+        } else {
+          // DIGIT1 is named "1", DIGIT2 is "2", etc.
+          tabIndex = Integer.valueOf(e.getCode().getName()) - 1;
         }
+        dashboard.selectTab(tabIndex);
```
AustinShalit: I think this might be more readable:
```
dashboard.selectTab(Integer.valueOf(e.getCode().getName().replace("Numpad ". "")) - 1);
```

nightpool: agreed, although the code is already merged. can you PR?

AustinShalit: Yea

AustinShalit: I was thinking about it for too long

PR #58, opened 2017-07-24
ScrollPane and height fixes for the gallery

### LINKED ISSUE #25: Widget palette needs a scrollbar if it doesn't all fit
When there are more widgets vertically in the palette than fit in the window there should be a scrollbar.

nightpool:
Now widgets fill the whole vbox, and they don't get super huge anymore. Also, the pane will scroll if the number of widgets is too large.

closes #25

### DIFF
--- a/src/main/java/edu/wpi/first/shuffleboard/components/WidgetGallery.java
+++ b/src/main/java/edu/wpi/first/shuffleboard/components/WidgetGallery.java
@@ -8,6 +8,7 @@ import javafx.beans.property.SimpleObjectProperty;
 import javafx.fxml.FXMLLoader;
 import javafx.scene.control.Label;
 import javafx.scene.layout.Pane;
+import javafx.scene.layout.Priority;
 import javafx.scene.layout.StackPane;
 import javafx.scene.layout.TilePane;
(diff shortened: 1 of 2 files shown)


AustinShalit:
The width of the pane is still a little large:
untitled

nightpool:
passes on linux, fails on mac

src/main/java/edu/wpi/first/shuffleboard/components/WidgetGallery.java
```diff
@@ -35,10 +36,10 @@ public WidgetGallery() {
    */
   public void loadWidgets(Collection<WidgetType> widgets) {
     widgets.stream().map(WidgetType::get)
-            .peek(widget -> {
+            .peek(widget ->
               DummySource.forTypes(widget.getDataTypes())
-                         .ifPresent(widget::setSource);
-            })
+                         .ifPresent(widget::setSource)
+            )
```
SamCarlberg: random style change?

nightpool: yes?

PR #65, opened 2017-07-28
Cleanup tab selection parsing

AustinShalit:
As discussed: https://github.com/wpilibsuite/shuffleboard/pull/57#pullrequestreview-52986476

### DIFF
--- a/src/main/java/edu/wpi/first/shuffleboard/MainWindowController.java
+++ b/src/main/java/edu/wpi/first/shuffleboard/MainWindowController.java
@@ -131,15 +131,11 @@ public class MainWindowController {
 
     root.addEventFilter(KeyEvent.KEY_PRESSED, e -> {
       if (e.isShortcutDown() && e.getCode().isDigitKey()) {
-        int tabIndex;
-        if (e.getCode().getName().contains("Numpad")) {
-          // remove leading "Numpad " to parse the number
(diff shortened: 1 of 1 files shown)


src/main/java/edu/wpi/first/shuffleboard/MainWindowController.java
```diff
@@ -131,15 +131,11 @@ private void initialize() throws IOException {
...
-        } else {
-          // DIGIT1 is named "1", DIGIT2 is "2", etc.
-          tabIndex = Integer.valueOf(e.getCode().getName()) - 1;
-        }
-        dashboard.selectTab(tabIndex);
+        /*
+         * Numpad digits have a name of "Numpad n"; where n is the number. We need to remove the
+         * leading "Numpad " to parse the number.  Digit keys do not have this issue.
+         */
+        dashboard.selectTab(Integer.valueOf(e.getCode().getName().replace("Numpad ", "")) - 1);
```
nightpool: Integer.valueOf(...) should be extracted into an intermediate variable for clarity.

AustinShalit: I disagree with extracting it to an intermediate variable.  I could put it on a new line.  Like this:
```
dashboard.selectTab(
    Integer.valueOf(e.getCode().getName().replace("Numpad ", "")) - 1);
```

PR #68, opened 2017-07-28
Update Checkstyle to 8.1

AustinShalit:
Also added some new checks.

### DIFF
--- a/src/main/java/edu/wpi/first/shuffleboard/components/WidgetPane.java
+++ b/src/main/java/edu/wpi/first/shuffleboard/components/WidgetPane.java
@@ -167,7 +167,7 @@ public class WidgetPane extends TilePane {
   }
 
   /**
-   * @return Returns the expected size of the widget, in tiles.
+   * Get the expected size of the widget, in tiles.
    */
   public TileSize sizeOfWidget(Widget widget) {
     Pane view = widget.getView();
--- a/src/main/java/edu/wpi/first/shuffleboard/widget/WidgetType.java
+++ b/src/main/java/edu/wpi/first/shuffleboard/widget/WidgetType.java
@@ -6,13 +6,14 @@ import java.util.Set;
 import java.util.function.Supplier;
 
 public interface WidgetType extends Supplier<Widget> {
+
   /**
-   * @return The name of the widget.
+   * Get the name of the widget (ex: "Number Slider").
    */
   String getName();
 
   /**
-   * @return The data types the widget should be suggested for.
+   * Get data types the widget should be suggested for.
    */
   Set<DataType> getDataTypes();
 }
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -46,7 +46,7 @@ dependencies {
 
 checkstyle {
     configFile = file("$rootDir/checkstyle.xml")
-    toolVersion = "6.19"
+    toolVersion = "8.1"
 }
 
 pmd {


checkstyle.xml
```diff
@@ -82,6 +82,11 @@
         <module name="ArrayTypeStyle"/>
         <module name="MissingSwitchDefault"/>
         <module name="FallThrough"/>
+        <module name="SimplifyBooleanExpression"/>
+        <module name="SimplifyBooleanReturn"/>
+        <module name="StringLiteralEquality"/>
+        <module name="UnnecessaryParentheses"/>
+        <module name="ExplicitInitializationCheck"/>
```
SamCarlberg: This was deliberately left out so we can be explicit with initializations. It's easy to forget that booleans default to `false`, for example.

AustinShalit: From: http://checkstyle.sourceforge.net/apidocs/com/puppycrawl/tools/checkstyle/checks/coding/ExplicitInitializationCheck.html

>Rationale: each instance variable gets initialized twice, to the same value. Java initializes each instance variable to its default value (0 or null) before performing any initialization specified in the code. So in this case, x gets initialized to 0 twice, and bar gets initialized to null twice. So there is a minor inefficiency. This style of coding is a hold-over from C/C++ style coding, and it shows that the developer isn't really confident that Java really initializes instance variables to default values.

SamCarlberg: >  So there is a minor inefficiency

Which is completely irrelevant. The code is more readable if you can go to the declaration and see what the initial value is *immediately* without having to remember or google what the default value is

src/main/java/edu/wpi/first/shuffleboard/sources/NetworkTableSource.java
```diff
@@ -17,7 +17,7 @@
 
   protected final String fullTableKey;
   private int listenerUid = -1;
-  private volatile boolean ntUpdate = false;
+  private volatile boolean ntUpdate;
```
nightpool: should be undone

src/main/java/edu/wpi/first/shuffleboard/sources/recording/Playback.java
```diff
@@ -41,7 +41,7 @@
   private final int numFrames;
   private final int maxFrameNum;
   private Thread autoRunner;
-  private volatile boolean started = false;
+  private volatile boolean started;
```
nightpool: ibid

src/main/java/edu/wpi/first/shuffleboard/widget/BooleanBox.java
```diff
@@ -43,7 +43,7 @@ private Color getColor() {
     final Boolean data = getData();
     if (data == null) {
       return Color.BLACK;
-    } else if (data == true) { //NOPMD
+    } else if (data) {
```
nightpool: was NOPMD for a reason

src/main/java/edu/wpi/first/shuffleboard/util/Maps.java
```diff
@@ -32,7 +32,7 @@ private Maps() {
    */
   public static final class MapBuilder<K, V> {
 
-    private boolean completed = false;
+    private boolean completed;
```
nightpool: same as above

src/main/java/edu/wpi/first/shuffleboard/widget/WidgetType.java
```diff
@@ -7,11 +7,15 @@
 
 public interface WidgetType extends Supplier<Widget> {
   /**
+   * Get the name of the widget.
+   *
```
nightpool: this is way too verbose, and just clutters up the code for no reason. Either remove the `@returns` statement or leave it as the only doc.

src/main/java/edu/wpi/first/shuffleboard/widget/WidgetType.java
```diff
@@ -7,11 +7,15 @@
...
    * @return The name of the widget.
    */
   String getName();
 
   /**
+   * Get the data types the widget should be suggested for.
+   *
```
nightpool: ibid

src/main/java/edu/wpi/first/shuffleboard/widget/Widgets.java
```diff
@@ -22,7 +22,7 @@
 public final class Widgets {
 
   private static Map<String, WidgetType> widgets = new TreeMap<>();
-  private static boolean loadedStockWidgets = false;
+  private static boolean loadedStockWidgets;
```
nightpool: booleans should be explicitly initialized

src/main/java/edu/wpi/first/shuffleboard/components/WidgetPane.java
```diff
@@ -151,7 +151,9 @@ public WidgetTile addWidget(Widget widget, GridPoint location, TileSize size) {
   }
 
   /**
-   * @return Returns the expected size of the widget, in tiles.
+   * Gets the expected size of the widget.
+   *
+   * @return the expected size of the widget, in tiles.
```
nightpool: redundant.

checkstyle.xml
nightpool: If we're not initializing an object value in the constructor, then that deserves to be called out and made explicit in the code using ` = null`. 

Booleans should *always* be explicitly initialized—the idea that one boolean type is "default" is an anachronistic holdover from C and not at all obvious to most people.

AustinShalit:
@nightpool Please review.

PR #69, opened 2017-07-28
Update to Gradle 4.0.2

AustinShalit:

PR #73, opened 2017-07-28
Rewrite build file to be strongly typed

JLLeitschuh:
This is built on top of #69

You can take it or leave this PR.
I just wanted to see how hard the exercise of rewriting the build to use the Gradle Kotlin DSL would be. Evidently it wasn't that hard. This took me ~25 min to do.


build.gradle.kts
```diff
@@ -0,0 +1,225 @@
...
+    mainClassName = "edu.wpi.first.shuffleboard.Shuffleboard"
+}
+
+val sourceJar = task<Jar>("sourceJar") {
+    description = "Creates a JAR that contains the source code."
+    from(java.sourceSets["main"].allSource)
+    classifier = "sources"
+}
+
+tasks.withType<Jar> {
```
JLLeitschuh: There used to be a `classifier=null` here when this was done using the shadow plugin. I don't know why that was there. I don't think it was strictly necessary.

SamCarlberg: Probably to remove the `-all` or `-shadow` suffix from the file name

build.gradle.kts
```diff
@@ -0,0 +1,225 @@
...
+    gradleVersion = "4.0.2"
+}
+
+/**
+ * Retrieves the [java][org.gradle.api.plugins.JavaPluginConvention] project convention.
+ */
+val Project.`java`: org.gradle.api.plugins.JavaPluginConvention get() =
+convention.getPluginByName<org.gradle.api.plugins.JavaPluginConvention>("java")
+
+
```
JLLeitschuh: Extra white space I should remove

build.gradle.kts
```diff
@@ -0,0 +1,221 @@
...
+/**
+ * Configures the [checkstyle][org.gradle.api.plugins.quality.CheckstyleExtension] project extension.
+ */
+fun Project.`checkstyle`(configure: org.gradle.api.plugins.quality.CheckstyleExtension.() -> Unit): Unit =
```
nightpool: Is the fully qualified package name necessary?

src/main/java/edu/wpi/first/shuffleboard/widget/NetworkTableTreeWidget.java
```diff
@@ -38,7 +39,7 @@
    * Compares tree items alphabetically.
    */
   private static final Comparator<TreeItem<NetworkTableEntry>> alphabetical
-      = Comparator.comparing(item -> item.getValue().getKey().toLowerCase());
+      = Comparator.comparing(item -> item.getValue().getKey().toLowerCase(Locale.getDefault()));
```
nightpool: unrelated change

build.gradle.kts
JLLeitschuh: This is what you get printed when you run the `gskGenAccessors` task. I just copied and pasted them. 

There's a different mode you can use that doesn't require these declarations in this file but I've never tried it. I could try it and see if it works.

build.gradle.kts
JLLeitschuh: Fixed, see line 150

gradle.properties
```diff
@@ -0,0 +1 @@
+org.gradle.script.lang.kotlin.accessors.auto=true
```
JLLeitschuh: This allowed me to not have all that extra fully classified cruft at the bottom of the `build.gradle.kts` file. Thanks @nightpool for suggesting that I clean that up or I wouldn't have tried this new config mode.

JLLeitschuh:
I cleaned this PR up quite a bit this morning. I re-added the shadow plugin. I tested the shadow JAR that the build produces and it runs shuffleboard.

I also added configuring the jar manifest in this PR so now the shadow jar will actually execute the main class when you run `java -jar shuffleboard.jar`. I don't know if this worked before but it definitely works now.

PR #83, opened 2017-08-02
Fix widgets not resizing in leftmost column

### LINKED ISSUE #76: Widgets only resize on the edge of the screen that is moving
If you have a row with 3 widgets that are {2, 1, 1} grid cells in size and the "2" sized widget could be shrunk. If I drag from the right edge towards the left edge, things only on the right edge start moving, then shrinking. The left edge widget doesn't seem to resize. Probably a bug since the other interior widgets seem to resize properly.

SamCarlberg:
Fixes #76


nightpool:
codecov, please

PR #86, opened 2017-08-02
Add a setting for allowing an indeterminate progress bar

### LINKED ISSUE #80: Showing negative numbers as progress bars doesn't work
Showing negative numbers as progress bars does the cylon eyes thing - very cool, but probably not what was expected.

AustinShalit:
closes #80 

ezgif-3-80740cfe4e

### DIFF
--- a/src/main/java/edu/wpi/first/shuffleboard/widget/ProgressBar.java
+++ b/src/main/java/edu/wpi/first/shuffleboard/widget/ProgressBar.java
@@ -4,7 +4,9 @@ import com.google.common.primitives.Doubles;
 import edu.wpi.first.shuffleboard.data.types.NumberType;
 
 import javafx.beans.binding.Bindings;
+import javafx.beans.property.BooleanProperty;
 import javafx.beans.property.DoubleProperty;
+import javafx.beans.property.SimpleBooleanProperty;
 import javafx.beans.property.SimpleDoubleProperty;
 import javafx.fxml.FXML;
 import javafx.scene.layout.Pane;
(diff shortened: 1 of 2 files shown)


SamCarlberg:
We should probably also make the default range (-1, 1) for consistency with the slider

src/main/java/edu/wpi/first/shuffleboard/widget/ProgressBar.java
```diff
@@ -19,12 +22,13 @@
 
   final DoubleProperty minValue = new SimpleDoubleProperty(this, "Min Value", 0);
   final DoubleProperty maxValue = new SimpleDoubleProperty(this, "Max Value", 1);
+  final BooleanProperty allowIndeterminateValue = new SimpleBooleanProperty(this, "Allow Indeterminate", false);
```
AustinShalit: I do not understand what change you need me to make.

AustinShalit: #80 claims that the default value should be false.  @bradamiller thoughts?

nightpool: allowIndeterminateValue -> allowIndeterminate

SamCarlberg: I was referring to the the string

nightpool:
To reiterate, having this as a setting doesn't make sense. it should just be the default behavior to constrain the value to a range.

AustinShalit:
@nightpool:
>This should at the very least be the default behavior.

What do you think should be the default behavior?

@SamCarlberg:
>We should probably also make the default range (-1, 1) for consistency with the slider

Good idea.  I will open a PR.

nightpool:
closed in favor of #88

PR #92, opened 2017-08-02
Add SimplifyBooleanExpression Checkstyle check

AustinShalit:


src/main/java/edu/wpi/first/shuffleboard/widget/BooleanBox.java
```diff
@@ -43,7 +43,7 @@ private Color getColor() {
     final Boolean data = getData();
     if (data == null) {
       return Color.BLACK;
-    } else if (data == true) { //NOPMD
```
nightpool: this was added specifically for readability. please revert it and disable the checkstyle check for this line.

AustinShalit: I just pushed a commit that checks the equality using the equals method.

nightpool: But it's still not consistent with the line above it?

AustinShalit: Updated again.  Take a look

nightpool:
Looks great!

PR #99, opened 2017-08-03
Refactor Widget class tree, add VoltageView Widget

### LINKED ISSUE #72: Tiles should use the "card" CSS class
This ties them in more with material design. We should also remove the blue background from tiles as part of this

SamCarlberg:
Adds a `LinearIndicator` component that looks like a progress bar, but doesn't have to start at the left edge.

Refactor the Widget class tree to allow annotated widgets with more than one type of data

Fix tiles not having their titles set when changing the source

Tiles now use the `card` CSS class; this closes #72

Screenshot of the new widget:

screenshot from 2017-08-04 13-37-42

### DIFF
--- a/src/main/java/edu/wpi/first/shuffleboard/DummySource.java
+++ b/src/main/java/edu/wpi/first/shuffleboard/DummySource.java
@@ -43,6 +43,9 @@ public class DummySource<T> extends AbstractDataSource<T> {
     } else if (types.contains(DataTypes.SendableChooser)) {
       final SendableChooserData data = new SendableChooserData(new String[]{"A", "B", "C"}, "A", "A");
       return Optional.of(new DummySource(DataTypes.SendableChooser, data));
+    } else if (types.stream().anyMatch(DataType::isComplex)) {
+      DataType type = types.stream().filter(DataType::isComplex).findFirst().get();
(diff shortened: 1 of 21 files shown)


nightpool:
Some usability things:

 - Voltage View should be -12 <-> 12 by default
 - we should also fix the general case of progress bars (so use linear indicators, but don't make them voltage views, I think?) Or maybe just unify voltage views with progress bars, where we do something special to show the units and change the default values if the type is an AnalogInput?
 - why can the user change the "center" value? Is there any situation where this shouldn't be zero? Since it doesn't actually change the center of the graph, if we still want it to be use-modifiable it should probably have a better name, like the "zeroing value"
 - Progress bars should *probably* automatically adopt their "default" values, if the user hasn't set them explicitly, to the max and min values ever seen. not exactly sure how to do this with just the property sheets/bindings, but maybe custom listeners would work?

nightpool:
Code wise, this refactor looks good. Usability wise I'm very happy, although there's still a small gap:

image

absolutely love the new card styles.

AustinShalit:
>Voltage View should be -12 <-> 12 by default

It is very common for voltages to be higher than 12v.

PR #98, opened 2017-08-03
Add AppVeyor Config

AustinShalit:

PR #100, opened 2017-08-04
Add SpeedController widget

SamCarlberg:
Automatically switches into a controllable mode when the LiveWindow is enabled; they do not switch back afterward.

Screenshot: Controllable on the left, view-only on the right

screenshot from 2017-08-04 16 37 55


nightpool:
what's going on with appveyor? cc @AustinShalit

AustinShalit:
I would merge #98 to fix the issue.

app/src/main/java/edu/wpi/first/shuffleboard/app/LiveWindow.java
```diff
@@ -0,0 +1,43 @@
...
+import javafx.beans.property.ReadOnlyBooleanProperty;
+import javafx.beans.property.SimpleBooleanProperty;
+
+/**
+ * Helper class for dealing with LiveWindow mode.
+ */
+public final class LiveWindow {
+
+  @VisibleForTesting
+  static ITable liveWindowTable = NetworkTable.getTable("LiveWindow");
```
nightpool: doesn't need to be visible

Issue #102, opened 2017-08-06
Should have support for custom plugins


SamCarlberg:
Plugins would have:

- A name
- A list of custom widgets
- A list of custom source types
- A list of custom data types

The name must always be set, but plugins can choose to define any or all of the custom lists.

We'll need to change `SourceType` away from an enum to an interface + util class, like `DataType`/`DataTypes`. Should probably also change `SourceType` to do data recording and playback, so we don't have to add custom code for each source type in the recording/playback classes.

We'd need to add UI stuff to load plugin jars or classes, as well as a simple dialog to show the currently loaded plugins.

Providing the plugin API would let us move classes into their own jars. All the stock widgets can be bundled into a `StockWidget` plugin, network table sources can be in a `NetworkTable Source` plugin, etc. Ideally the stock data types would also go into a plugin, but that would mean plugins would either need to be compiled with that jar to let the `@Description` annotation to function, forcing us to load that jar first to make sure the data type classes are loaded; or remove the `@Description` annotation entirely.

Because of this, I think we should keep the stock data types in the api jar (#101) to make sure all plugins can use all the data types provided by WPILib, but still split the stock widgets and data sources into plugins.

There are a few pain points like the network table tree widget that we use to display data in network tables and add widgets from. We may want to let source types provide a JavaFX node that we stick in a tab in the sources view.

nightpool:
closed by #140

PR #104, opened 2017-08-07
Splits app into API and APP project

### LINKED ISSUE #101: Should publish standalone API jar
A tentative list of the classes that should be moved into an `api` project:

| Package | Class |
|---|---|
| `components` | `LinearIndicator` 
|| `NumberField` |
||`Scrubber`|
| `data` | Everything in this package |
| `data.types` | Everything in this package |
| `properties` | `AsyncProperty` |
| `sources` | `DataSource` |
| | `AbstractDataSource` |
| | `SourceType` (should also be split into an interface and util class) |
| | `IncompatibleSourceException` |
| `widget` | `Widget`
| | `AbstractWidget` |
| | `AnnotatedWidget` |
| | `SingleTypeWidget` |
| | `ComplexAnnotatedWidget` |
| | `@Description` |
| | `@ParamtrizedController`

Other thoughts:

- Should this be a fatjar or standalone? A fat jar would give users access to guava etc. but would increase the jar size 
- Need to look into how plugins would interact with an API jar and determine any extra classes that should get moved over. Particularly custom JavaFX components.

@JLLeitschuh can you work on splitting the project into an `api` project and an `app` project?

JLLeitschuh:
Closes #101

This splits the project into two separate projects. One called `api` and the other called `app`. In the future the `api` will be published as an API that teams can write plugins against.


settings.gradle
```diff
@@ -0,0 +1,12 @@
...
+include ":app"
+
+rootProject.children.each { project ->
+    final String groovyName = "${project.name}.gradle"
+    final String kotlinName = "${project.name}.gradle.kts"
+    project.buildFileName = groovyName
+    if (!project.buildFile.isFile()) {
+        project.buildFileName = kotlinName
+    }
+    assert project.buildFile.isFile() : "File named $groovyName or $kotlinName must exist."
```
JLLeitschuh: Fancy renaming, so instead of every file being named `build.gradle.kts` instead its `${project.name}.gradle.kts`. Much nicer for searching for in your IDE.

api/api.gradle.kts
```diff
@@ -0,0 +1,16 @@
+plugins {
+    `java-library`
```
JLLeitschuh: https://docs.gradle.org/current/userguide/java_library_plugin.html

settings.gradle
SamCarlberg: You should put that in a comment in the file

ThadHouse: I like this. Would this be something we would be interested in doing in the other repositories?

JLLeitschuh: Are you proposing to use more Gradle Kotlin DSL stuff? You're going to have to have that conversation with the greater team?
Or are you just referring to having the gradle build file names start with the name of the project?

ThadHouse: Just the build file names. I do agree about keeping all the builds consistent with groovy.

JLLeitschuh:
There are failing network tables tests that I'm not sure how to fix. @SamCarlberg @nightpool?

api/src/main/java/edu/wpi/first/shuffleboard/api/sources/SingleKeyNetworkTableSource.java
```diff
@@ -1,8 +1,8 @@
-package edu.wpi.first.shuffleboard.sources;
+package edu.wpi.first.shuffleboard.api.sources;
```
SamCarlberg: This class isn't part of the API

api/src/main/java/edu/wpi/first/shuffleboard/api/sources/NetworkTableSource.java
```diff
@@ -1,8 +1,8 @@
-package edu.wpi.first.shuffleboard.sources;
+package edu.wpi.first.shuffleboard.api.sources;
```
SamCarlberg: This class isn't part of the API

api/src/main/java/edu/wpi/first/shuffleboard/api/sources/CompositeNetworkTableSource.java
```diff
@@ -1,8 +1,8 @@
-package edu.wpi.first.shuffleboard.sources;
+package edu.wpi.first.shuffleboard.api.sources;
```
SamCarlberg: This class isn't part of the API

api/src/test/java/edu/wpi/first/shuffleboard/api/sources/CompositeNetworkTableSourceTest.java
```diff
@@ -1,18 +1,18 @@
-package edu.wpi.first.shuffleboard.sources;
+package edu.wpi.first.shuffleboard.api.sources;
```
SamCarlberg: `CompositeNetworkTableSource` should be in `:app`, so this should be moved as well

api/src/test/java/edu/wpi/first/shuffleboard/api/sources/SingleKeyNetworkTableSourceTest.java
```diff
@@ -1,10 +1,10 @@
-package edu.wpi.first.shuffleboard.sources;
+package edu.wpi.first.shuffleboard.api.sources;
```
SamCarlberg: `SingleKeyNetworkTableSource` should be in `:app`, so this should be moved as well

api/src/main/java/edu/wpi/first/shuffleboard/api/sources/NetworkTableSource.java
JLLeitschuh: It is, `SourceType` depends upon it.

JLLeitschuh:
There are several classes that were moved around because things you wanted in the `API` project had transitive dependencies.
I offer two proposals:
1. Fix these problems in this PR by pushing more changes to my branch.
2. Merge this branch and then fix the problems of having too much in the `API` project.

api/src/main/java/edu/wpi/first/shuffleboard/api/sources/SingleKeyNetworkTableSource.java
JLLeitschuh: `NetworkTablesSource` depends upon this.

api/src/main/java/edu/wpi/first/shuffleboard/api/sources/NetworkTableSource.java
SamCarlberg: I mentioned in #101 that `SourceType` should be split into an interface and util class, like we have for `DataType`. The interface and util class would live in `:api`, and the implementations would be in `:app`. None/Static should still be defined in the util class IMO

api/src/test/java/edu/wpi/first/shuffleboard/api/sources/CompositeNetworkTableSourceTest.java
JLLeitschuh: `NetworkTableSource` depends upon `CompositeNetworkTableSource`.

SamCarlberg:
It should be addressed as part of this PR

api/src/test/java/edu/wpi/first/shuffleboard/api/sources/SingleKeyNetworkTableSourceTest.java
JLLeitschuh: See: https://github.com/wpilibsuite/shuffleboard/pull/104#discussion_r131720009

app/src/main/java/edu/wpi/first/shuffleboard/app/json/JsonBuilder.java
```diff
@@ -1,4 +1,4 @@
-package edu.wpi.first.shuffleboard.json;
+package edu.wpi.first.shuffleboard.app.json;
```
nightpool: Line 19 needs to be fixed

SamCarlberg: Fixed with 6f7fbe2

app/src/main/java/edu/wpi/first/shuffleboard/app/sources/NetworkTableSourceType.java
```diff
@@ -0,0 +1,18 @@
+package edu.wpi.first.shuffleboard.app.sources;
+
+import edu.wpi.first.shuffleboard.api.sources.SourceType;
+import edu.wpi.first.shuffleboard.api.sources.SourceTypes;
+
+public final class NetworkTableSourceType extends SourceType {
```
SamCarlberg: @nightpool thoughts on this?

api/src/main/java/edu/wpi/first/shuffleboard/api/sources/SourceType.java
```diff
@@ -1,38 +1,40 @@
...
-  CAMERA_SERVER(true, "camera_server://", __ -> {
-    throw new UnsupportedOperationException("Not implemented");
-  });
+public class SourceType {
 
-  public final boolean isRecordable;
+  private final String name;
+  private final boolean isRecordable;
   private final String protocol;
   private final Function<String, DataSource> sourceSupplier;
```
nightpool: This should be a method now that it's not an enum

SamCarlberg: IMO it's nice to b able to just call the constructor and pass it the function it needs, rather than making it an abstract method and forcing anonymous classes or a subclass for each type

nightpool: yes but see ^^ about annotations.

app/src/main/java/edu/wpi/first/shuffleboard/app/sources/NetworkTableSourceType.java
JLLeitschuh: Gotta love those startup static initializer race conditions.

SamCarlberg: `DataType` doesn't use annotations. I think this is fine, especially since it'll get changed (again) when we add plugins for #102

settings.gradle
JLLeitschuh: Comment added per @SamCarlberg's request.

PR #107, opened 2017-08-07
Gradle 4.1

JLLeitschuh:
https://docs.gradle.org/4.1/release-notes.html


build.gradle.kts
```diff
@@ -33,7 +33,7 @@ subprojects {
     }
 
     dependencies {
-        testCompile(group = "junit", name = "junit", version = "+")
+        "testCompile"(create(group = "junit", name = "junit", version = "+"))
```
JLLeitschuh: 😢

api/api.gradle.kts
```diff
@@ -7,8 +7,6 @@ Public API for writing plugins for shuffleboard.
 """.trimMargin()
 
 dependencies {
-    fun api(group: String, name: String, version: String?, classifier : String? = null) =
-        add("api", create(group = group, name = name, version = version, classifier = classifier))
```
JLLeitschuh: 😄

app/app.gradle.kts
```diff
@@ -1,7 +1,5 @@
 import com.github.jengelman.gradle.plugins.shadow.tasks.ShadowJar
 import org.gradle.jvm.tasks.Jar
-import org.gradle.script.lang.kotlin.create
-import org.gradle.script.lang.kotlin.invoke
```
JLLeitschuh: 👍

nightpool:
project-schema is generate right? also, can you link the release notes for 4.1 in the body of this PR for posterities sake?

gradle/README.md
```diff
@@ -0,0 +1,6 @@
+# Gradle
+
+To regenerate the [project-schema.json](project-schema.json) file run this in the project's root directory:
+```bash
+./gradlew kotlinDslAccessorsSnapshot
```
nightpool: This doesn't explain when or why you would do this though, or what the project-schema document is....

JLLeitschuh: You want me to put something in about it?

From what I know it's simply gradle way of figuring out what all of the extensions are after all of the plugins have settled and fully defined all of their configurations.

nightpool: Should it be tracked in source or not? I can't find any documentation on it at all.

build.gradle.kts
JLLeitschuh: I don't think these are coming back either. It will only be accessible in sub project `*.gradle.kts` files where the plugin is fully applied, not in the `subprojects` block.

PR #110, opened 2017-08-08
Add simple encoder widget

SamCarlberg:
Just shows the speed and distance in two text fields


AustinShalit:
Sorry that I am behind on reviewing code.

This PR works great for encoders that are being used to measure distance.  What about encoders that are being used to measure an angle?

SamCarlberg:
Encoders only report rate, distance, and distance per tick

```java
allwpilib/wpilibj/src/athena/java/edu/wpi/first/wpilibj/Encoder.java

  @Override
  public void updateTable() {
    if (m_table != null) {
      m_table.putNumber("Speed", getRate());
      m_table.putNumber("Distance", getDistance());
      m_table.putNumber("Distance per Tick", EncoderJNI.getEncoderDistancePerPulse(m_encoder));
    }
  }
```

nightpool:
Austin, that sounds like a good change to make on the WPILib end.

PR #111, opened 2017-08-08
Switches shuffleboard to new ntcore artifact locations

ThadHouse:
Required fixing NetworkTableUtils, as the exception throwing functions
were removed.

closes #112

### DIFF
--- a/api/api.gradle.kts
+++ b/api/api.gradle.kts
@@ -10,5 +10,6 @@ dependencies {
     api(group = "com.google.guava", name = "guava", version = "21.0")
     api(group = "org.fxmisc.easybind", name = "easybind", version = "1.0.3")
     api(group = "org.controlsfx", name = "controlsfx", version = "8.40.11")
-    api(group = "edu.wpi.first.wpilib.networktables.java", name = "NetworkTables", version = "+", classifier = "desktop")
+    api(group = "edu.wpi.first.ntcore", name = "ntcore-java", version = "3.1.7-20170808143930-12-gccfeab5")
+    implementation(group = "edu.wpi.first.wpiutil", name = "wpiutil-java", version = "2.0.0-20170808143537-16-gf0cc5d9")
 }
--- a/api/src/main/java/edu/wpi/first/shuffleboard/api/util/NetworkTableUtils.java
+++ b/api/src/main/java/edu/wpi/first/shuffleboard/api/util/NetworkTableUtils.java
@@ -132,7 +132,7 @@ public final class NetworkTableUtils {
       return DataTypes.Map;
     }
     if (rootTable.containsKey(normalKey)) {
-      return DataType.forJavaType(rootTable.getValue(normalKey).getClass());
+      return DataType.forJavaType(rootTable.getValue(normalKey, null).getClass());
     }
     if (rootTable.containsSubTable(normalKey)) {
       ITable table = rootTable.getSubTable(normalKey);
--- a/app/app.gradle.kts
+++ b/app/app.gradle.kts
@@ -18,6 +18,7 @@ dependencies {
     compile(group = "com.google.code.gson", name = "gson", version = "2.8.1")
     fun testFx(name: String, version: String = "4.0.+") =
         create(group = "org.testfx", name = name, version = version)
+    runtime(group = "edu.wpi.first.ntcore", name = "ntcore-jni", version = "3.1.7-20170808143930-12-gccfeab5", classifier = "all")
     testCompile(testFx(name = "testfx-core"))
     testCompile(testFx(name = "testfx-junit"))
     testRuntime(testFx(name = "openjfx-monocle", version = "8u76-b04"))


app/app.gradle.kts
```diff
@@ -18,6 +18,7 @@ dependencies {
     compile(group = "com.google.code.gson", name = "gson", version = "2.8.1")
     fun testFx(name: String, version: String = "4.0.+") =
         create(group = "org.testfx", name = name, version = version)
+    runtime(group = "edu.wpi.first.ntcore", name = "ntcore-jni", version = "+", classifier = "all")
```
SamCarlberg: Why `+` instead of a fixed version?

ThadHouse: The old one was plus, and we don't have a release version yet. It's all development.

JLLeitschuh: The problem with using a plus though is that if you push a version that breaks something then the shuffleboard CI servers will start failing for an unrelated reason. This should probably be locked to a specific version, especially if it's all development releases.

api/api.gradle.kts
```diff
@@ -10,5 +10,6 @@ dependencies {
     api(group = "com.google.guava", name = "guava", version = "21.0")
     api(group = "org.fxmisc.easybind", name = "easybind", version = "1.0.3")
     api(group = "org.controlsfx", name = "controlsfx", version = "8.40.11")
-    api(group = "edu.wpi.first.wpilib.networktables.java", name = "NetworkTables", version = "+", classifier = "desktop")
-}
\ No newline at end of file
+    api(group = "edu.wpi.first.ntcore", name = "ntcore-java", version = "+")
```
JLLeitschuh: @SamCarlberg and @nightpool 
Just curious, do you really want network tables as part of the public API of shuffleboard?
If not, you just want it to be an internal implementation detail then this should use the `implementation` configuration instead.

api/api.gradle.kts
```diff
@@ -10,5 +10,6 @@ dependencies {
...
+    api(group = "edu.wpi.first.wpiutil", name = "wpiutil-java", version = "+")
```
JLLeitschuh: Is this a dependency of `ntcore` already?
If so is it really a public part of the API that shuffleboard wants to expose?

app/app.gradle.kts
ThadHouse: What worries me about that is I get it, but our release script on Jenkins depends on plus. We would have to change our entire release cycle just to have this be static. None of our other repos use ststic, and never have.

api/api.gradle.kts
ThadHouse: The transitive dependencies are not working right because of the way the publishing pipeline works. So we do need both. And eventually, plugins might want to use it, as we might add actual functionality. Same with keeping ntcore as part of API.

app/app.gradle.kts
JLLeitschuh: > None of our other repos use static, and never have.

That's not true:
https://github.com/WPIRoboticsProjects/GRIP/blob/master/build.gradle#L254-L255

ThadHouse: None of our other projects published with WPILib in the plugins.

ThadHouse: GRIP isn't build with the plugins, and is on its own release schedule. I feel like this will end up getting published with the plugins, and I don't know if anyone will have time to change the release pipeline just to support a static version before beta.

api/api.gradle.kts
JLLeitschuh: So there's a problem in the POM file that network tables publishes?
Do you have an issue open to track that?

ThadHouse: Its actually much deeper then that. It's actually an issue that spans across multiple repos. It's an artifact from needing the combined jars to be published all at the same time.

api/api.gradle.kts
nightpool: yes it is currently part of the public API.

api/api.gradle.kts
nightpool: We don't need to be depending on wpiutil, to the best of my understanding.

nightpool: We don't need to be depending on wpiutil, to the best of my understanding. this can be removed.

api/src/main/java/edu/wpi/first/shuffleboard/api/util/NetworkTableUtils.java
```diff
@@ -132,7 +132,7 @@ public static DataType dataTypeForEntry(String key) {
       return DataTypes.Map;
     }
     if (rootTable.containsKey(normalKey)) {
-      return DataType.forJavaType(rootTable.getValue(normalKey).getClass());
+      return DataType.forJavaType(rootTable.getValue(normalKey, null).getClass());
```
nightpool: what is the value of using this if we're already checking containsKey?

app/app.gradle.kts
nightpool: this should at least be pinned to v3 so we don't unexpectedly get breaking changes.

I don't understand how the release pipeline versioning affects libraries like ntcore. What is the concern? that we're going to be building with a different version then it was designed for? That's *already* a problem with the current setup. 

Anyway, if it's not currently getting built with the plugins, then we don't need to worry about this until it does, no?

api/src/main/java/edu/wpi/first/shuffleboard/api/util/NetworkTableUtils.java
SamCarlberg: The exception-throwing version of these methods (with only 1 argument) were removed

api/api.gradle.kts
ThadHouse: We do now. The new ntcore now have a java wpiutil dependency. Eventually we will be adding more utilities available for downstream projects.

JLLeitschuh: If this doesn't need to be a part of the public api of shuffleboard then this should be labeled as an `implementation` dependency.

api/src/main/java/edu/wpi/first/shuffleboard/api/util/NetworkTableUtils.java
ThadHouse: It would have thrown an exception any time this returns null. Because of the containsKey, this should be fine (other then not thread safe and potentially now has a race condition)

app/app.gradle.kts
ThadHouse: We don't have a V3 in this location. This location only has development releases.

ThadHouse: http://first.wpi.edu/FRC/roborio/maven/development/edu/wpi/first/ntcore/ntcore-java/3.1.7-20170808143930-12-gccfeab5/

This would be the version we would have to put in manually now.

ThadHouse: Also we have a whole release pipeline. What it does is start from the bottom (wpiutil now), and builds/publishes it. Then builds the next layer up, then up then up. These are versions that are published to release rather then development, so the versions are completely different, even though they have the same as the latest development release.

api/api.gradle.kts
```diff
@@ -10,5 +10,6 @@ dependencies {
     api(group = "com.google.guava", name = "guava", version = "21.0")
     api(group = "org.fxmisc.easybind", name = "easybind", version = "1.0.3")
     api(group = "org.controlsfx", name = "controlsfx", version = "8.40.11")
-    api(group = "edu.wpi.first.wpilib.networktables.java", name = "NetworkTables", version = "+", classifier = "desktop")
-}
\ No newline at end of file
+    api(group = "edu.wpi.first.ntcore", name = "ntcore-java", version = "3.1.7-20170808143930-12-gccfeab5")
+    api(group = "edu.wpi.first.wpiutil", name = "wpiutil-java", version = "2.0.0-20170808143537-16-gf0cc5d9")
```
JLLeitschuh: This should be an `implementation` dependency.

ThadHouse: Still disagree on that, because it might be usable for plugins later, and seems silly to have to go and forget later, then wonder why for some reason code doesn't work.

ThadHouse: But whatever, its your decision.

PR #113, opened 2017-08-08
Add tab preferences and autopopulation

SamCarlberg:
Adds "SmartDashboard" and "LiveWindow" default tabs that autopopulate on `"/SmartDashboard/"` and `"/LiveWindow/"`, respectively.

Autopopulation is smart and only creates a widget for a source if that source _and_ a parent source are not already present. For example, if `/LiveWindow/Motor/` is present, a widget for `/LiveWindow/Motor/Value` will not be added

Properties like tile size, hgap, and vgap are also configurable

screenshot from 2017-08-09 13 11 16

---

Known issues

- [x] Preferred widget sizes are not respected during autopopulation
  - This leads to some funky stuff with encoder widgets extending past the tile bounds
- [x] Widget panes default to 1x1 until they become visible, so the number of rows and columns are manually set when creating a tab
- [x] Autopopulation triggers when the source prefix changes; typing in the textfield to change it essentially triggers one population run per key stroke


333fred:
I'd suggest that Hgap and Vgap aren't great names, prefer using something Horizontal Spacing.

JLLeitschuh:
10% drop in code coverage is pretty significant.

JLLeitschuh:
A new photo when you fix the Hgap and Vgap thing would be nice.

api/src/main/java/edu/wpi/first/shuffleboard/api/sources/SourceType.java
```diff
@@ -61,4 +64,11 @@ public String removeProtocol(String text) {
     return sourceSupplier.apply(removeProtocol(uri));
   }
 
+  /**
+   * Gets a list of the IDs of all available sources of this type.
+   */
+  public ObservableList<String> getAvailableSourceIds() {
```
nightpool: 1. What's a source ID? sources don't have IDs, they have URIs
2. this should be more explicit what "available" means (all currently known sources) and that this is only some subset of all sources and may not be implemented for every type

app/src/main/java/edu/wpi/first/shuffleboard/app/components/DashboardTabPane.java
```diff
@@ -102,7 +136,96 @@ public DashboardTab(String title) {
...
+        }
+      }
+    }
+
+    private boolean shouldAutopopulate(String sourceId) {
+      return sourceId.startsWith(getSourcePrefix())
+          || SourceTypes.stripProtocol(sourceId).startsWith(getSourcePrefix());
+    }
+
+    private boolean isNotMetadata(String sourceId) {
```
nightpool: here too

app/src/main/java/edu/wpi/first/shuffleboard/app/sources/NetworkTableSourceType.java
```diff
@@ -11,8 +16,27 @@
...
     super("NetworkTable", true, "network_table://", NetworkTableSource::forKey);
+    NetworkTablesJNI.addEntryListener("", (uid, key, value, flags) -> {
+      NetworkTableUtils.getHierarchy(key)
+          .stream()
+          .map(this::toUri)
+          .forEach(uri -> {
+            if (NetworkTableUtils.isDelete(flags)) {
+              availableSourceIds.remove(uri);
+            } else if (!availableSourceIds.contains(uri)) {
+              availableSourceIds.add(uri);
```
nightpool: is this thread-safe? should be on a runLater if not

nightpool:
mostly good, some nits

api/src/main/java/edu/wpi/first/shuffleboard/api/util/NetworkTableUtils.java
```diff
@@ -119,6 +119,14 @@ public static boolean isDelete(int flags) {
   }
 
   /**
+   * Checks if the given key is metadata, eg matches the format "~METADATA~" or ".metadata"
+   */
+  public static boolean isMetadata(String key) {
```
JLLeitschuh: Simple thing to unit test. You should probably add some for this.

api/src/main/java/edu/wpi/first/shuffleboard/api/util/NetworkTableUtils.java
```diff
@@ -119,6 +119,14 @@ public static boolean isDelete(int flags) {
...
+    return key.matches("^.*/\\..+(/.*)?$")
```
SamCarlberg: I should precompile this as a `Pattern`, actually. But yeah, the regex could be cleaner

api/src/test/java/edu/wpi/first/shuffleboard/api/util/PropertyUtilsTest.java
```diff
@@ -0,0 +1,42 @@
...
+    assertEquals(3, combine.size());
+    b.add("B");
+    assertEquals(4, combine.size());
+    assertEquals(a.get(1), combine.get(2));
+    assertEquals(b.get(1), combine.get(3));
+
+    a.clear();
+    b.clear();
+    assertEquals(0, combine.size());
+  }
```
JLLeitschuh: What happens when you call `combine.clear()`? Do both `a` and `b` get cleared too?

nightpool: this binding is entirely one-way, like all javafx bindings.

app/src/main/java/edu/wpi/first/shuffleboard/app/components/DashboardTabPane.java
```diff
@@ -102,7 +136,96 @@ public DashboardTab(String title) {
...
+            Widgets.createWidget(Widgets.widgetNamesForSource(source).get(0), source)
+                .ifPresent(w -> getWidgetPane().addWidget(w));
+          }
+        }
+      }
+    }
+
+    private boolean shouldAutopopulate(String sourceId) {
+      return sourceId.startsWith(getSourcePrefix())
+          || SourceTypes.stripProtocol(sourceId).startsWith(getSourcePrefix());
```
SamCarlberg: Sorry, holdover from recording/playback

api/src/main/java/edu/wpi/first/shuffleboard/api/util/NetworkTableUtils.java
AustinShalit: https://github.com/wpilibsuite/OutlineViewer/blob/JavaFX/src/test/java/edu/wpi/first/outlineviewer/model/MetadataTest.java

JLLeitschuh: So copy pasta the test from outline viewer?

AustinShalit: @PeterJohnson What do you think about moving this method into either ntcore or some other common tool jar?

SamCarlberg: It's not an ntcore thing, it's specific to LiveWindow

AustinShalit: Maybe we should create a LiveWindow API Jar

nightpool:
Our metadata detection us different from yours though. I don't think every
five like method needs a new library—especially not on Java.

AustinShalit:
It does not need to be a library, but it would be nice say when we want to change what makes metadata from `~DATA~` to `.data`.

api/src/main/java/edu/wpi/first/shuffleboard/api/sources/SourceType.java
SamCarlberg: > What's a source ID? sources don't have IDs, they have URIs

`Source.getId()`. URI is a notion of SourceType.

api/src/test/java/edu/wpi/first/shuffleboard/api/util/PropertyUtilsTest.java
```diff
@@ -0,0 +1,42 @@
...
+    ObservableList<String> a = FXCollections.observableArrayList("a");
+    ObservableList<String> b = FXCollections.observableArrayList("b");
+    ObservableList<String> combine = PropertyUtils.combineLists(a, b);
+
+    a.add("A");
+    assertEquals(3, combine.size());
+    b.add("B");
+    assertEquals(4, combine.size());
+    assertEquals(a.get(1), combine.get(2));
+    assertEquals(b.get(1), combine.get(3));
```
SamCarlberg: This is the ordering I would expect

app/src/main/java/edu/wpi/first/shuffleboard/app/components/DashboardTabPane.java
SamCarlberg: Actually, `Source` has a `getId()` method. This matches up with that API

app/src/main/java/edu/wpi/first/shuffleboard/app/components/DashboardTabPane.java
```diff
@@ -102,7 +136,96 @@ public DashboardTab(String title) {
...
+        });
+        dialog.showAndWait();
+      })));
+    }
+
+    private void populate() {
+      if (getWidgetPane().getScene() == null || getWidgetPane().getParent() == null) {
+        // Defer until the pane is visible and is laid out in the scene
+        deferPopulation = true;
+        Platform.runLater(this::populate);
```
SamCarlberg: I'm not very happy with this. I was counting 3000+ deferrals per startup... not fun.

api/src/main/java/edu/wpi/first/shuffleboard/api/sources/SourceTypes.java
```diff
@@ -1,19 +1,39 @@
...
+
+import javafx.beans.InvalidationListener;
+import javafx.collections.FXCollections;
+import javafx.collections.ObservableList;
 
 public final class SourceTypes {
 
   private static final Map<String, SourceType> types = new HashMap<>();
+  private static final ObservableList<String> typeNames = FXCollections.observableArrayList();
+  private static final ObservableList<String> namesUnmodifiable = FXCollections.unmodifiableObservableList(typeNames);
```
nightpool: does this really need to be a static variable? how often do we call getNames?

nightpool: seems unused, actually.

api/src/main/java/edu/wpi/first/shuffleboard/api/sources/SourceTypes.java
```diff
@@ -68,4 +84,39 @@ public static SourceType forName(String name) {
...
+   * @param uri the uri to strip the protocol from
+   */
+  public static String stripProtocol(String uri) {
+    return typeForUri(uri).removeProtocol(uri);
+  }
+
+  /**
+   * Gets a read-only observable list of the names of the registered source types.
+   */
+  public static ObservableList<String> getNames() {
```
nightpool: unused?

api/src/main/java/edu/wpi/first/shuffleboard/api/util/NetworkTableUtils.java
```diff
@@ -21,6 +22,9 @@
    */
   public static final ITable rootTable = NetworkTable.getTable("");
 
+  private static final Pattern oldMetadataPattern = Pattern.compile("/~\\w+~/?");
+  private static final Pattern newMetadataPattern = Pattern.compile("/\\.");
```
nightpool: This should be `/(\\.|~[^/]+~/)`

PR #116, opened 2017-08-09
Set default widgets for data types

### LINKED ISSUE #114: There should be a default widget for each data type
We currently just grab the first compatible widget alphabetically. We should have an annotation `@DefaultWidget` that provides an array of data types (in the same format as `@Description`) that can be placed on widget classes

```java
@Description(name  = "My Number Widget", dataTypes = NumberType.class)
@DefaultWidget(NumberType.class)
public class MyNumberWidget implements Widget {
  // ...
}
```

SamCarlberg:
Provide a `@DefaultWidgetFor` annotation that is used to set a default widget for certain data types. If no default widget is set, falls back to use the first alphabetical widget.

Closes #114


JLLeitschuh:
If you are fixing a bug you should probably write a test so that it doesn't regress.
This isn't a UI bug, so the tests shouldn't be flaky.

AustinShalit:
I don't agree with this solution. Widgets should not know if they are the default data type for a source. It should be the other way around. Sources should have a default widget.

SamCarlberg:
It should not be source-level at all. It should belong to either widgets or data types. Putting it on widgets lets third parties override the default widgets we provide with their own just by adding an annotation to their widget classes. Otherwise, they would need to call (for example)

```java
class MyWidget implements Widget {

  static {
    DataTypes.Foo.setDefaultWidget(MyWidget.class);
  }

}
```

AustinShalit:
Where in our code right now is the prioritization for widgets outside of our code?

Also, what happens if 2 widgets are annotated as the default for the same type?

SamCarlberg:
> Where in our code right now is the prioritization for widgets outside of our code?

Nowhere. It's something we have to keep in mind while designing our APIs, however.

> Also, what happens if 2 widgets are annotated as the default for the same type?
 
The most recently registered widget will be used.

JLLeitschuh:
> > Also, what happens if 2 widgets are annotated as the default for the same type?
> 
> The most recently registered widget will be used.

Sounds like you want some sort of priority. That way a plugin can annotate itself with a higher precedence priority than the default and be the default.

AustinShalit:
appveyor is failing due to #117

app/src/main/java/edu/wpi/first/shuffleboard/app/widget/StockWidgets.java
```diff
@@ -30,6 +33,14 @@ public static void init() {
...
                .filter(c -> c.isAnnotationPresent(Description.class))
                .forEach(Widgets::register);
+      Widgets.setDefaultWidget(DataTypes.String, "Text View");
+      Widgets.setDefaultWidget(DataTypes.Boolean, "Boolean Box");
+      Widgets.setDefaultWidget(DataTypes.Number, "Number Slider");
+      Widgets.setDefaultWidget(DataTypes.Map, "Network Table Tree");
+      Widgets.setDefaultWidget(DataTypes.SendableChooser, "ComboBox Chooser");
+      Widgets.setDefaultWidget(DataTypes.Encoder, "Encoder");
+      Widgets.setDefaultWidget(DataTypes.AnalogInput, "Voltage View");
+      Widgets.setDefaultWidget(DataTypes.SpeedController, "Speed Controller");
```
AustinShalit: We should use the overload of this method that takes a WidgetType rather than a String

SamCarlberg: The `WidgetType` instances are created by `Widgets.register`, we don't have the handles. It's a non-issue; this is just temporary until plugins are added

AustinShalit: We are depending on Strings to enact behavior.  What is going to be changed with the plugins?

SamCarlberg: They'll be able to specify which widgets to use as defaults for which data types as part of the API

AustinShalit: How are they going to specify that?  With Strings?

SamCarlberg: `Map<DataType, WidgetType>`. Tentatively. I haven't hammered out a full specification yet

AustinShalit: Ok

JLLeitschuh:
Can you change the name of this PR to something more descriptive?

PR #118, opened 2017-08-09
Update waitForNtcoreEvents to fix race

### LINKED ISSUE #117: waitForNtcoreEvents() is broken
There is a race with the waitForNtcoreEvents method.  Take a look at this example code:

```
@Test
public void testDelete() {
  table.putString("testDelete", "value");
  // Listener could be fired here before we wait for a NetworkTable listener to be called!
  waitForNtcoreEvents();
  // ...
}

### PR DESCRIPTION
**Pull request by AustinShalit**
closes #117

### DIFF
--- a/api/src/main/java/edu/wpi/first/shuffleboard/api/util/NetworkTableUtils.java
+++ b/api/src/main/java/edu/wpi/first/shuffleboard/api/util/NetworkTableUtils.java
@@ -151,10 +151,24 @@ public final class NetworkTableUtils {
    */
   public static void waitForNtcoreEvents() {
     CompletableFuture<?> future = new CompletableFuture<>();
-    NetworkTablesJNI.addEntryListener("", (uid, key, value, flags) -> {
-      future.complete(null);
+    final String indexKey = "waitForNtcoreEvents";
+
+    NetworkTablesJNI.addEntryListener(indexKey, (uid, key, value, flags) -> {
+      NetworkTablesJNI.deleteEntry(indexKey);
       NetworkTablesJNI.removeEntryListener(uid);
-    }, 0xFF);
+    }, ITable.NOTIFY_LOCAL);
+    NetworkTablesJNI.addEntryListener(indexKey, (uid, key, value, flags) -> {
+      NetworkTablesJNI.removeEntryListener(uid);
+      future.complete(null);
+    }, ITable.NOTIFY_LOCAL & ITable.NOTIFY_DELETE);
+
+    /*
+     * This works because all notifications are put into a single queue and are processed by a
+     * single thread.
+     *
+     * https://github.com/wpilibsuite/shuffleboard/pull/118#issuecomment-321374691
+     */
+    NetworkTablesJNI.putBoolean(indexKey, false);
     future.join();
   }
 
--- a/api/src/main/java/edu/wpi/first/shuffleboard/api/util/NetworkTableUtils.java
+++ b/api/src/main/java/edu/wpi/first/shuffleboard/api/util/NetworkTableUtils.java
@@ -156,11 +156,11 @@ public final class NetworkTableUtils {
     NetworkTablesJNI.addEntryListener(indexKey, (uid, key, value, flags) -> {
       NetworkTablesJNI.deleteEntry(indexKey);
       NetworkTablesJNI.removeEntryListener(uid);
-    }, ITable.NOTIFY_LOCAL);
+    },  ITable.NOTIFY_NEW | ITable.NOTIFY_LOCAL);
     NetworkTablesJNI.addEntryListener(indexKey, (uid, key, value, flags) -> {
       NetworkTablesJNI.removeEntryListener(uid);
       future.complete(null);
-    }, ITable.NOTIFY_LOCAL & ITable.NOTIFY_DELETE);
+    }, ITable.NOTIFY_DELETE | ITable.NOTIFY_LOCAL);
 
     /*
      * This works because all notifications are put into a single queue and are processed by a
--- a/app/src/test/java/edu/wpi/first/shuffleboard/app/json/WidgetSaverTest.java
+++ b/app/src/test/java/edu/wpi/first/shuffleboard/app/json/WidgetSaverTest.java
@@ -10,7 +10,6 @@ import edu.wpi.first.wpilibj.networktables.NetworkTablesJNI;
 import javafx.stage.Stage;
 import org.junit.After;
 import org.junit.Before;
-import org.junit.Ignore;
 import org.junit.Test;
 import org.testfx.framework.junit.ApplicationTest;
 
@@ -46,7 +45,6 @@ public class WidgetSaverTest extends ApplicationTest {
             .getValue();
   }
 
-  @Ignore
   @Test
   public void loadSimpleWidget() throws Exception {
     String widgetJson = "{\n"
--- a/app/src/test/java/edu/wpi/first/shuffleboard/app/json/WidgetSaverTest.java
+++ b/app/src/test/java/edu/wpi/first/shuffleboard/app/json/WidgetSaverTest.java
@@ -10,6 +10,7 @@ import edu.wpi.first.wpilibj.networktables.NetworkTablesJNI;
 import javafx.stage.Stage;
 import org.junit.After;
 import org.junit.Before;
+import org.junit.Ignore;
 import org.junit.Test;
 import org.testfx.framework.junit.ApplicationTest;
 
@@ -45,6 +46,7 @@ public class WidgetSaverTest extends ApplicationTest {
             .getValue();
   }
 
+  @Ignore
   @Test
   public void loadSimpleWidget() throws Exception {
     String widgetJson = "{\n"
--- a/app/src/test/java/edu/wpi/first/shuffleboard/app/sources/CompositeNetworkTableSourceTest.java
+++ b/app/src/test/java/edu/wpi/first/shuffleboard/app/sources/CompositeNetworkTableSourceTest.java
@@ -9,7 +9,6 @@ import edu.wpi.first.wpilibj.networktables.NetworkTable;
 
 import org.junit.After;
 import org.junit.Before;
-import org.junit.Ignore;
 import org.junit.Test;
 
 import static edu.wpi.first.shuffleboard.api.util.NetworkTableUtils.waitForNtcoreEvents;
@@ -39,7 +38,6 @@ public class CompositeNetworkTableSourceTest {
     assertTrue(source.getData().isEmpty());
   }
 
-  @Ignore
   @Test
   public void testDataUpdates() {
     CompositeNetworkTableSource<MapData> source

### CODE REVIEW

File: api/src/main/java/edu/wpi/first/shuffleboard/api/util/NetworkTableUtils.java [hunk 1]
```diff
@@ -148,14 +148,18 @@ public static DataType dataTypeForEntry(String key) {
...
+  public static void waitForNtcoreEvents(Runnable runnable) {
     CompletableFuture<?> future = new CompletableFuture<>();
-    NetworkTablesJNI.addEntryListener("", (uid, key, value, flags) -> {
+    int listener = NetworkTablesJNI.addEntryListener("", (uid, key, value, flags) -> {
       future.complete(null);
       NetworkTablesJNI.removeEntryListener(uid);
     }, 0xFF);
+    runnable.run();
     future.join();
+    NetworkTablesJNI.removeEntryListener(listener);
```
nightpool: duplicate of line 156?

app/src/test/java/edu/wpi/first/shuffleboard/app/components/NetworkTableTreeTest.java
```diff
@@ -99,12 +97,11 @@ public void testBranches() {
...
-    table.putString("sub_a/sub_entry_a", "");
-    table.putString("a", "");
-    table.putString("b", "");
-    table.putString("sub_b/sub_entry_b", "");
-    waitForNtcoreEvents();
+    waitForNtcoreEvents(() -> table.putString("c", ""));
+    waitForNtcoreEvents(() -> table.putString("sub_a/sub_entry_a", ""));
+    waitForNtcoreEvents(() -> table.putString("a", ""));
+    waitForNtcoreEvents(() -> table.putString("b", ""));
+    waitForNtcoreEvents(() -> table.putString("sub_b/sub_entry_b", ""));
```
nightpool: these should all be in the same waitFor call

app/src/test/java/edu/wpi/first/shuffleboard/app/json/DashboardTabPaneSaverTest.java
```diff
@@ -29,15 +29,21 @@ public void setUp() {
...
+        -> NetworkTablesJNI.putDouble("/LiveWindow/Elevator/p", 0.5));
+    NetworkTableUtils.waitForNtcoreEvents(()
+        -> NetworkTablesJNI.putDouble("/LiveWindow/Elevator/d", 0.5));
+    NetworkTableUtils.waitForNtcoreEvents(()
+        -> NetworkTablesJNI.putDouble("/LiveWindow/Elevator/f", 0.5));
+
+    NetworkTableUtils.waitForNtcoreEvents(()
+        -> NetworkTablesJNI.putBoolean("/LiveWindow/~STATUS~/LW Enabled", true));
+    NetworkTableUtils.waitForNtcoreEvents(()
+        -> NetworkTablesJNI.putBoolean("/LiveWindow/Elevator/enabled", false));
```
nightpool: ibid

api/src/main/java/edu/wpi/first/shuffleboard/api/util/NetworkTableUtils.java
```diff
@@ -148,14 +148,18 @@ public static DataType dataTypeForEntry(String key) {
...
    */
-  public static void waitForNtcoreEvents() {
+  public static void waitForNtcoreEvents(Runnable runnable) {
     CompletableFuture<?> future = new CompletableFuture<>();
-    NetworkTablesJNI.addEntryListener("", (uid, key, value, flags) -> {
+    int listener = NetworkTablesJNI.addEntryListener("", (uid, key, value, flags) -> {
       future.complete(null);
       NetworkTablesJNI.removeEntryListener(uid);
     }, 0xFF);
+    runnable.run();
```
JLLeitschuh: Couldn't the listener potentially get called before this runnable gets called? Thus it would be complete before network tables gets the value?

nightpool:
Can you also try removing the @Disabled tests and seeing if this fixes them?

app/src/test/java/edu/wpi/first/shuffleboard/app/json/DashboardTabPaneSaverTest.java
JLLeitschuh: "ibid"? Explain?

nightpool: same comment as the one preceding this one (https://en.wikipedia.org/wiki/Ibid.)

SamCarlberg: "As above". Apparently it's latin

api/src/main/java/edu/wpi/first/shuffleboard/api/util/NetworkTableUtils.java
AustinShalit: Yes.  Missed that.  Fixed

app/src/test/java/edu/wpi/first/shuffleboard/app/components/NetworkTableTreeTest.java
AustinShalit: They cannot be do to limitations of the way the code is written.

app/src/test/java/edu/wpi/first/shuffleboard/app/json/DashboardTabPaneSaverTest.java
AustinShalit: See above.

app/src/test/java/edu/wpi/first/shuffleboard/app/components/NetworkTableTreeTest.java
nightpool: whats the problem with:

```java
waitForNtcoreEvents(() -> {
  table.putString("c", "");
  table.putString("sub_a/sub_entry_a", "");
  table.putString("a", "");
  table.putString("b", "");
  table.putString("sub_b/sub_entry_b", "");
});
```

api/src/main/java/edu/wpi/first/shuffleboard/api/util/NetworkTableUtils.java
nightpool: Yes, this is still a problem.

api/src/main/java/edu/wpi/first/shuffleboard/api/util/NetworkTableUtils.java
```diff
@@ -148,13 +148,16 @@ public static DataType dataTypeForEntry(String key) {
...
+   * @param runnable A single NetworkTable call to wait on.
    */
-  public static void waitForNtcoreEvents() {
+  public static void waitForNtcoreEvents(Runnable runnable) {
     CompletableFuture<?> future = new CompletableFuture<>();
     NetworkTablesJNI.addEntryListener("", (uid, key, value, flags) -> {
       future.complete(null);
       NetworkTablesJNI.removeEntryListener(uid);
     }, 0xFF);
+    runnable.run();
```
nightpool: re-upping with @JLLeitschuh's concern from an earlier commit:
> Couldn't the listener potentially get called before this runnable gets called? Thus it would be complete before network tables gets the value?

nightpool: This problem seems intractable to me if we're using a live networktable.

app/src/test/java/edu/wpi/first/shuffleboard/app/components/NetworkTableTreeTest.java
AustinShalit: The future will complete when the listener for `table.putString("c", "");` is called but not for `table.putString("sub_b/sub_entry_b", "");`

AustinShalit:
What if we do something like this: @JLLeitschuh @nightpool 
```
  public static void waitForNtcoreEvents(Runnable runnable) {
    CompletableFuture<?> future = new CompletableFuture<>();

    NetworkTablesJNI.addEntryListener("waitForNtcoreEvents", (uid, key, value, flags) -> {
      NetworkTablesJNI.deleteEntry("waitForNtcoreEvents");
      NetworkTablesJNI.removeEntryListener(uid);
    }, ITable.NOTIFY_LOCAL);
    NetworkTablesJNI.addEntryListener("waitForNtcoreEvents", (uid, key, value, flags) -> {
      NetworkTablesJNI.removeEntryListener(uid);
      future.complete(null);
    }, ITable.NOTIFY_LOCAL);
    
    runnable.run();
    NetworkTablesJNI.putBoolean("waitForNtcoreEvents", false);
    future.join();
  }
```

api/src/main/java/edu/wpi/first/shuffleboard/api/util/NetworkTableUtils.java
AustinShalit: See my code example below.  Only call the listener when it has the NOTIFY_LOCAL flag.

nightpool:
1. I'm not sure the second listener is buying us much, since we don't care about the deletion being synchronous, unless i'm misunderstanding

2. This assumes all network table updates are all put onto a strictly-ordered queue and done in the same thread. This seems like a very strong assumption. Is it true? Is it something we can assume will  be true in the future?

app/src/test/java/edu/wpi/first/shuffleboard/app/components/NetworkTableTreeTest.java
AustinShalit: My code example below might solve this.

AustinShalit:
>I'm not sure the second listener is buying us much, since we don't care about the deletion being synchronous, unless I'm misunderstanding

We need to guarantee that the value is deleted before we return so it does not interfere with tests.

>This assumes all network table updates are all put onto a strictly-ordered queue and done in the same thread. This seems like a very strong assumption. Is it true? Is it something we can assume will be true in the future?

I think this is the case but @PeterJohnson will know.

api/src/main/java/edu/wpi/first/shuffleboard/api/util/NetworkTableUtils.java
JLLeitschuh: > This problem seems intractable to me if we're using a live networktable.

You could figure it out but you would need to compare the key and value and make sure that the event matches the change.

PeterJohnson:
> This approach assumes all network table updates are all put onto a strictly-ordered queue and done in the same thread. This seems like a very strong assumption. Is it true? Is it something we can assume will be true in the future?

Yes, that's correct, and I don't expect it to change in the future.  All notification events are put into a single queue and processed by a single thread (see https://github.com/wpilibsuite/ntcore/blob/master/src/main/native/cpp/Notifier.cpp#L126).  All user callbacks operate out of the same thread (note this does mean that user callbacks will block each other).  The only exception is when we add multiple instances--in that case each instance will have its own notifier thread.

nightpool:
Okay, LGTM then % removing the @Ignored for some of the currently-disabled tests

JLLeitschuh:
The chunk of code in this comment should really be put in a network tables testlib that ships as an independent JAR.
https://github.com/wpilibsuite/shuffleboard/pull/118#issuecomment-321341629

That is some complex threading logic but looks like it would work.

AustinShalit:
>Okay, LGTM then % removing the @Ignored for some of the currently-disabled tests

There was one test that this change should fix.  I removed the annotation from it.

>The chunk of code in this comment should really be put in a network tables testlib that ships as an independent JAR.

Maybe something for another day.  Maybe we should open an issue on ntcore to have a way of waiting for the queue to clear rather than saving this code.

>Please make sure you put a comment with @PeterJohnson's explanation and a link to this conversation for documentation purposes.

Done

PR #119, opened 2017-08-09
Reduce Comment Size from codecov

AustinShalit:
Depends on #118 

I do not know why disabling codecov is not working right now.  This PR will reduce the size of codecov comments to make them less obtrusive.  

Please wait to merge until codecov comments on this PR so we can confirm that we like it.

### DIFF
--- a/codecov.yml
+++ b/codecov.yml
@@ -12,4 +12,7 @@ coverage:
     patch: yes
     changes: no
 
-comment: false
+comment:
+  layout: "diff"
+  behavior: new
+  require_changes: yes


SamCarlberg:
Please rebase onto `master` instead of `fix117`

AustinShalit:
I can't.  Master does not pass all CI.  codecov will not comment unless CI passes.

JLLeitschuh:
> Please rebase onto master instead of fix117

> Depends on #118

codecov.yml
```diff
@@ -5,11 +5,14 @@ codecov:
 coverage:
   precision: 2
   round: down
-  range: "70...100"
+  range: "60...100"
```
nightpool: this change seems unrelated and also not great.

AustinShalit: Changed back to 70

JLLeitschuh: I think that saying that code coverage isn't important is like saying, you enjoy writing legacy code.

> Legacy code is code without tests
> \- Michael Feathers' Working Effectively with Legacy Code (ISBN 0-13-117705-2)

Theres two sides to this, you aren't writing tests for yourself, you are writing tests for the next developer that comes along and is trying to add a new feature without breaking existing functionality.

PR #139, opened 2017-08-11
Only allow a single instance of shuffleboard to launch

AustinShalit:


app/src/main/java/edu/wpi/first/shuffleboard/app/Shuffleboard.java
```diff
@@ -40,12 +42,21 @@ public void init() throws Exception {
...
+      });
+
+      primaryStage.setScene(new Scene(mainPane));
+      primaryStage.setMinWidth(640);
+      primaryStage.setMinHeight(480);
+      primaryStage.setWidth(Screen.getPrimary().getVisualBounds().getWidth());
+      primaryStage.setHeight(Screen.getPrimary().getVisualBounds().getHeight());
+      primaryStage.show();
+    } catch (AlreadyLockedException e) {
+      JUnique.sendMessage("alreadyRunning", "true");
```
nightpool: sendMessage takes appId, which we're passing earlier as `getClass().getCanonicalName()`

nightpool:
Why choose start over init? Shouldn't we abandon as early as possible if we know we don't need to load resources/layouts/etc?

app/src/main/java/edu/wpi/first/shuffleboard/app/Shuffleboard.java
AustinShalit: Right

AustinShalit:
>Why choose start over init? Shouldn't we abandon as early as possible if we know we don't need to load resources/layouts/etc?

I want the functionality of bringing the primary stage of the application that was already open to the front.  That object does not exist until start has been called.

nightpool:
hmm. isn't that just

```java 
(message -> {
    if (mainPane != null) {
        mainPane.getStage().bringToFront()
    }
})
```

?

AustinShalit:
Updated.

PR #140, opened 2017-08-11
Add plugin support

SamCarlberg:
Moves builtin widgets and data types to `:plugins:base`

Moves network table sources and widgets to `:plugins:networktables`

Autogenerates source trees for each registered source type

### Features
- [x] Exposes a plugin API (see [Plugin.java](https://github.com/wpilibsuite/shuffleboard/pull/140/files#diff-6f25aca9b3741be4f1c8ff86f315a2dc))
  - [x] Allow custom widgets
  - [x] Allow custom source types
  - [x] Allow custom data types
  - [x] Allow setting default widgets on a per-data-type basis
  - [x] Allow custom type adapters for recording/playback
  - [x] Define custom data playback in `SourceType`; nothing is hard-coded anymore in `:api` or `:app`
  - [x] Expose custom names, descriptions, and version numbers
  - [x] Expose custom themes
- [x] Allow plugins to be dynamically loaded an unloaded at runtime, not just startup
  - (This has a very nice dialog, see below)
  - [x] Remove widgets when the plugins defining them are unloaded
  - [x] Remove widgets sources when plugins defining those sources or data types are unloaded
  - [x] Restore sources when the relevant plugins are reloaded
- [x] Read JAR files and load all plugins found
- [ ] Read .class files and load all plugins found

### Known issues
- [x] Startup takes about 5 seconds to load the UI due to LiveWindow tab autopopulation

#### Plugin preferences dialog
Plugin preferences dialog


SamCarlberg:
Dialog for loading/unloading plugins

screenshot from 2017-08-11 14 53 55

api/api.gradle.kts
```diff
@@ -12,4 +12,11 @@ dependencies {
     api(group = "org.controlsfx", name = "controlsfx", version = "8.40.11")
     api(group = "edu.wpi.first.ntcore", name = "ntcore-java", version = "3.1.7-20170808143930-12-gccfeab5")
     implementation(group = "edu.wpi.first.wpiutil", name = "wpiutil-java", version = "2.0.0-20170808143537-16-gf0cc5d9")
+    runtime(group = "edu.wpi.first.ntcore", name = "ntcore-jni", version = "3.1.7-20170808143930-12-gccfeab5", classifier = "all")
+    testRuntime(group = "edu.wpi.first.ntcore", name = "ntcore-jni", version = "3.1.7-20170808143930-12-gccfeab5", classifier = "all")
+    fun testFx(name: String, version: String = "4.0.+") =
+            create(group = "org.testfx", name = name, version = version)
+    testCompile(testFx(name = "testfx-core"))
+    testCompile(testFx(name = "testfx-junit5"))
```
JLLeitschuh: Since this `testCompile` dependency is shared now in both `api` and `app` you should put it there so you don't have duplicate version numbers.

SamCarlberg: "There" being...?

api/src/main/java/edu/wpi/first/shuffleboard/api/data/DataType.java
```diff
@@ -9,52 +11,58 @@
...
 
   /**
    * Gets the name of this data type.
    */
-  String getName();
+  public final String getName() {
+    return name;
+  }
+
+  public final Class<T> getJavaClass() {
```
JLLeitschuh: Vs getting the Groovy class or the Kotlin class?

api/src/main/java/edu/wpi/first/shuffleboard/api/data/DataType.java
```diff
@@ -9,52 +11,58 @@
...
   }
 
-  /**
-   * Gets the data type with the given name.
-   */
-  static DataType<?> forName(String name) {
-    return DataTypes.forName(name).orElse(DataTypes.Unknown);
+  @Override
+  public int hashCode() {
+    return getClass().hashCode();
```
JLLeitschuh: Why aren't you also hashing on the name?

api/src/main/java/edu/wpi/first/shuffleboard/api/data/DataType.java
```diff
@@ -9,52 +11,58 @@
...
-      return DataTypes.BooleanArray;
-    } else if (type == byte[].class || type == Byte[].class) {
-      return DataTypes.RawBytes;
-    } else {
-      return DataTypes.Unknown;
+  @Override
+  public boolean equals(Object obj) {
+    if (obj == null) {
+      return false;
     }
```
JLLeitschuh: Usually, you have a check for this immediately after the null check as an optimization:
```java
if (this == obj) return true;
```

api/src/main/java/edu/wpi/first/shuffleboard/api/data/DataTypes.java
```diff
@@ -86,6 +61,104 @@ public static void register(DataType<?> dataType) {
...
+        List<DataType> sorted = dataTypes.values().stream()
+            .filter(t -> t != All)
+            .sorted((t1, t2) -> classComparator.reversed().compare(t1.getJavaClass(), t2.getJavaClass()))
+            .collect(Collectors.toList());
+        return Optional.of(sorted.get(0));
+      }
+    });
+  }
+
+  private static Comparator<Class<?>> closestTo(Class<?> target) {
```
JLLeitschuh: Can you explain why this is needed and what it is? Also, this is a complex bit of logic, are you sure that Guava doesn't have some sort of reflection utility to do this?
https://github.com/google/guava/wiki/ReflectionExplained

Also, if you do need it, I really hope that this is unit tested.

api/src/main/java/edu/wpi/first/shuffleboard/api/data/DataTypes.java
```diff
@@ -86,6 +61,104 @@ public static void register(DataType<?> dataType) {
...
+   */
+  public static Set<DataType> forJavaTypes(Class<?>... types) {
+    return Stream.of(types)
+        .map(DataTypes::forJavaType)
+        .filter(Optional::isPresent)
+        .map(Optional::get)
+        .collect(Collectors.toSet());
+  }
+
+  private static Class<?> boxedType(Class<?> primitiveType) {
```
JLLeitschuh: This could be replaced with `Primitives.wrap`:

http://google.github.io/guava/releases/21.0/api/docs/com/google/common/primitives/Primitives.html#wrap-java.lang.Class-

app/src/main/java/edu/wpi/first/shuffleboard/app/MainWindowController.java
```diff
@@ -85,48 +97,27 @@ private void initialize() throws IOException {
...
+    PluginLoader.getDefault().getKnownPlugins().addListener((ListChangeListener<Plugin>) c -> {
+      while (c.next()) {
+        if (c.wasAdded()) {
+          c.getAddedSubList().forEach(plugin -> {
+            plugin.loadedProperty().addListener((__, was, is) -> {
+              if (is) {
+                System.out.println("GENERATING FOR " + plugin.getName());
+                setup(plugin);
+              } else {
+                System.out.println("REMOVING ITEMS FOR " + plugin.getName());
```
JLLeitschuh: You probably want to move to using a logger like slf4j

app/src/main/java/edu/wpi/first/shuffleboard/app/MainWindowController.java
```diff
@@ -213,7 +271,7 @@ public void save() {
   private void saveAs() {
     FileChooser chooser = new FileChooser();
     chooser.getExtensionFilters().setAll(
-            new FileChooser.ExtensionFilter("SmartDashboard Save File (.json)", "*.json"));
+        new FileChooser.ExtensionFilter("SmartDashboard Save File (.json)", "*.json"));
```
JLLeitschuh: Formatter changes?

app/src/main/java/edu/wpi/first/shuffleboard/app/MainWindowController.java
```diff
@@ -251,7 +309,7 @@ public void load() {
     FileChooser chooser = new FileChooser();
     chooser.setInitialDirectory(new File(Storage.STORAGE_DIR));
     chooser.getExtensionFilters().setAll(
-            new FileChooser.ExtensionFilter("SmartDashboard Save File (.json)", "*.json"));
+        new FileChooser.ExtensionFilter("SmartDashboard Save File (.json)", "*.json"));
```
JLLeitschuh: Formatter changes?

app/src/main/java/edu/wpi/first/shuffleboard/app/plugin/PluginLoader.java
```diff
@@ -0,0 +1,124 @@
...
+import javafx.collections.FXCollections;
+import javafx.collections.ObservableList;
+
+public class PluginLoader {
+
+  private static final PluginLoader defaultLoader = new PluginLoader();
+
+  private final ObservableList<Plugin> knownPlugins = FXCollections.observableArrayList();
+
+  public static PluginLoader getDefault() {
```
JLLeitschuh: Public singletons make things hard to unit test with fakes. Avoid if possible.

app/src/main/java/edu/wpi/first/shuffleboard/app/PluginPaneController.java
```diff
@@ -0,0 +1,82 @@
...
+  @FXML
+  private TableView<Plugin> pluginTable;
+
+  private TableColumn<Plugin, String> nameColumn = new TableColumn<>("Plugin");
+  private TableColumn<Plugin, Boolean> loadedColumn = new TableColumn<>("Loaded");
+
+  @FXML
+  private void initialize() {
+    nameColumn.setCellValueFactory(param -> new ReadOnlyStringWrapper(param.getValue().getName()));
+    final ObservableList<Plugin> loadedPlugins = PluginLoader.getDefault().getKnownPlugins();
```
JLLeitschuh: You should probably pass in the `pluginLoader` instead of pulling it from a static context.

app/src/main/java/edu/wpi/first/shuffleboard/app/PluginPaneController.java
```diff
@@ -0,0 +1,82 @@
...
+    loadedColumn.setCellValueFactory(param -> {
+      Plugin plugin = param.getValue();
+      SimpleBooleanProperty prop = new SimpleBooleanProperty(true);
+      prop.addListener((__, was, is) -> {
+        if (was) {
+          PluginLoader.getDefault().unload(plugin);
```
JLLeitschuh: Avoid using a static singleton.

app/src/main/java/edu/wpi/first/shuffleboard/app/PluginPaneController.java
```diff
@@ -0,0 +1,82 @@
...
+        } else {
+          PluginLoader.getDefault().load(plugin);
```
JLLeitschuh: Avoid using a static singleton.

app/src/main/java/edu/wpi/first/shuffleboard/app/PluginPaneController.java
```diff
@@ -0,0 +1,82 @@
...
+    }
+    files.forEach(f -> {
+      if (f.getName().endsWith(".class")) {
+        System.out.println("LOAD CLASS FILE");
+        //TODO
+      } else if (f.getName().endsWith(".jar")) {
+        System.out.println("LOAD JAR FILE");
+        try {
+          JarFile jar = new JarFile(f);
+          PluginLoader.getDefault().loadPluginJar(jar);
```
JLLeitschuh: Avoid using a static singleton.

app/src/main/java/edu/wpi/first/shuffleboard/app/plugin/PluginLoader.java
```diff
@@ -0,0 +1,124 @@
...
+              .filter(p -> p.getClass().getName().equals(clazz.getName()))
+              .forEach(this::unload);
+          Plugin plugin = (Plugin) clazz.newInstance();
+          load(plugin);
+          return true;
+        }
+      } catch (ReflectiveOperationException e) {
+        // TODO log it?
+        //e.printStackTrace();
+        return false;
```
JLLeitschuh: Yes, this should definitely be logged.

app/src/main/java/edu/wpi/first/shuffleboard/app/plugin/PluginLoader.java
```diff
@@ -0,0 +1,124 @@
...
+      }
+    }
+    return false;
```
JLLeitschuh: Log in this case too.

app/src/test/java/edu/wpi/first/shuffleboard/app/json/DashboardTabPaneSaverTest.java
```diff
@@ -19,11 +19,12 @@
 
 import static org.junit.jupiter.api.Assertions.assertEquals;
 
+@Disabled
```
JLLeitschuh: Why is this whole test disabled?

plugins/networktables/networktables.gradle.kts
```diff
@@ -0,0 +1,3 @@
+dependencies {
```
JLLeitschuh: Above this line add a `description = "[SOME HELPFUL DESCRIPTION HERE]"`

plugins/base/base.gradle.kts
```diff
@@ -0,0 +1,11 @@
+plugins {
+    `java-library`
+}
+
+description = """
+Base shuffleboard plugin that provides the default data types and widgets.
+""".trimMargin()
+
+dependencies {
+    foo
```
JLLeitschuh: Wat?

plugins/plugins.gradle.kts
```diff
@@ -0,0 +1,14 @@
...
+        plugin("com.github.johnrengelman.shadow")
+        plugin("java-library")
+    }
+    dependencies {
+        compile(project(":api"))
+        fun testFx(name: String, version: String = "4.0.+") =
+                create(group = "org.testfx", name = name, version = version)
+        testCompile(testFx(name = "testfx-core"))
+        testCompile(testFx(name = "testfx-junit5"))
+        testRuntime(testFx(name = "openjfx-monocle", version = "8u76-b04"))
```
JLLeitschuh: This code is again duplicated here.

api/api.gradle.kts
JLLeitschuh: `build.gradle.kts` in the root project.

app/src/main/java/edu/wpi/first/shuffleboard/app/MainWindowController.java
SamCarlberg: I left those in while I was debugging. We use Java's builtin `Logger` class for logging; this was a quick and dirty (and easy to read) way to debug

app/src/main/java/edu/wpi/first/shuffleboard/app/plugin/PluginLoader.java
SamCarlberg: That'll be logged a lot; one statement per non-plugin class. For example, the `base` plugin jar would log ~20 times here

app/src/test/java/edu/wpi/first/shuffleboard/app/json/DashboardTabPaneSaverTest.java
SamCarlberg: Dependent on network tables

plugins/base/base.gradle.kts
SamCarlberg: Oops, was trying to see why it wasn't evaluating declarations like `compile` and `testRuntime`

api/src/main/java/edu/wpi/first/shuffleboard/api/data/DataType.java
SamCarlberg: I should really stop using IntelliJ's autogenerated equals/hashcode methods, they miss stuff like this

app/src/main/java/edu/wpi/first/shuffleboard/app/PluginPaneController.java
SamCarlberg: We'd need to set up a DI framework for that. Maybe later when we're near feature complete

SamCarlberg:
## Plugin window screenshot
screenshot from 2017-08-13 22 56 37

api/api.gradle.kts
```diff
@@ -12,4 +12,6 @@ dependencies {
     api(group = "org.controlsfx", name = "controlsfx", version = "8.40.11")
     api(group = "edu.wpi.first.ntcore", name = "ntcore-java", version = "3.1.7-20170808143930-12-gccfeab5")
     implementation(group = "edu.wpi.first.wpiutil", name = "wpiutil-java", version = "2.0.0-20170808143537-16-gf0cc5d9")
+    runtime(group = "edu.wpi.first.ntcore", name = "ntcore-jni", version = "3.1.7-20170808143930-12-gccfeab5", classifier = "all")
+    testRuntime(group = "edu.wpi.first.ntcore", name = "ntcore-jni", version = "3.1.7-20170808143930-12-gccfeab5", classifier = "all")
```
AustinShalit: Do you need this line?

api/src/main/java/edu/wpi/first/shuffleboard/api/data/DataType.java
```diff
@@ -9,52 +11,70 @@
...
+  public abstract boolean isComplex();
+
+  @Override
+  public boolean equals(Object obj) {
+    if (obj == null) {
+      return false;
+    }
+    if (this == obj) {
+      return true;
+    }
```
AustinShalit: `this == obj` should be first.

api/src/main/java/edu/wpi/first/shuffleboard/api/plugin/Plugin.java
```diff
@@ -0,0 +1,123 @@
...
+import java.util.List;
+
+import javafx.beans.property.BooleanProperty;
+import javafx.beans.property.SimpleBooleanProperty;
+
+/**
+ * Common superclass for plugins that can be loaded by the app at startup or during runtime. Subclasses must have
+ * public no-arg constructor or they will not be loaded.
+ */
+public class Plugin {
```
AustinShalit: abstract?

app/src/test/java/edu/wpi/first/shuffleboard/app/json/WidgetSaverTest.java
```diff
@@ -22,7 +21,7 @@
 
   @Override
   public void start(Stage stage) throws Exception {
-    Widgets.discover();
+    // Just here so can run on the FX thread
```
AustinShalit: Seems weird that we have code that needs to run on the FX Thread without using a stage.  Not something to fix in this PR but we should look into that.

api/src/main/java/edu/wpi/first/shuffleboard/api/plugin/Plugin.java
SamCarlberg: This makes it easy to create new instances for testing. There are also no abstract methods deliberately because implementations may implement any or none of the methods of this class

api/src/main/java/edu/wpi/first/shuffleboard/api/plugin/Plugin.java
```diff
@@ -117,15 +126,29 @@ public void onUnload() {
     return ImmutableList.of();
   }
 
-  public boolean isLoaded() {
+  /**
+   * Gets a list of themes that this plugin defines.
+   */
+  public List<Theme> getThemes() {
```
JLLeitschuh: Should the return type of this method be `ImmutableList<Theme>`?

api/src/main/java/edu/wpi/first/shuffleboard/api/util/Debouncer.java
```diff
@@ -0,0 +1,61 @@
...
+  private final long debounceDelay;
+  private final TimeUnit unit;
+
+  /**
+   * Creates a new debouncer.
+   *
+   * @param debounceDelay the maximum time delta between calls that should be allowed
+   * @param unit          the unit of {@code debounceDelay}
+   */
+  public Debouncer(long debounceDelay, TimeUnit unit) {
```
JLLeitschuh: You could replace these two args with with `java.time.Duration`

api/src/main/java/edu/wpi/first/shuffleboard/api/util/Debouncer.java
```diff
@@ -0,0 +1,61 @@
...
+  /**
+   * Debounces a function call. Generally, this should only be used to debounce multiple calls to the same method/lambda
+   * expression.
+   *
+   * @param runnable the functional unit to be run
+   */
+  public void debounce(Runnable runnable) {
+    if (future != null && !future.isDone()) {
+      future.cancel(false);
+    }
```
JLLeitschuh: So this schedules the runnable again after canceling the previous call to the `runnable`? What if the previous `runnable` was about to complete? It sounds like this isn't debouncing.

SamCarlberg: Say the debounce delay is 100ms. Calling debounce once will schedule the runnable to run after 100ms have passed. Calling it again before the delay is up will cancel a non-running future. Calling it after 100ms have passed should not cancel it, because the specified delay has passed without being interrupted by another call

api/src/main/java/edu/wpi/first/shuffleboard/api/util/Registry.java
```diff
@@ -57,6 +59,23 @@ public final boolean isRegistered(T item) {
...
+   * Registers many items at once. This is equivalent to
+   * <pre><code>
+   *   for (T item : items) {
+   *     register(item);
+   *   }
+   * </code></pre>
+   *
+   * @param items the items to register
+   */
+  public final void registerAll(T... items) {
```
JLLeitschuh: Annotate with `@SafeVarargs`?

app/src/main/resources/edu/wpi/first/shuffleboard/app/MainWindow.fxml
```diff
@@ -22,6 +23,8 @@
                 <MenuItem onAction="#load" text="Load layout" accelerator="SHORTCUT+O"/>
                 <MenuItem onAction="#showPrefs" text="Preferences" accelerator="SHORTCUT+,"/>
                 <SeparatorMenuItem/>
+                <MenuItem onAction="#showPlugins" text="Plugins" accelerator="SHORTCUT+P"/>
```
AustinShalit: This could be confusing because shortcut+p is usually reserved for printing.

JLLeitschuh: I agree!

api/src/main/java/edu/wpi/first/shuffleboard/api/data/DataType.java
JLLeitschuh: I agree

api/src/main/java/edu/wpi/first/shuffleboard/api/util/Debouncer.java
JLLeitschuh: Can you write a unit test that actually demonstrates what you are trying to do here. Its not at all clear the behaviour you are expecting out of this class. I think that "Debouncing" is not the correct term but I can't exactly tell because I don't see any tests that show the expected behaviour.

app/src/main/resources/edu/wpi/first/shuffleboard/app/MainWindow.fxml
SamCarlberg: I had considered `CTRL+L`, that may be better

api/src/main/java/edu/wpi/first/shuffleboard/api/util/Debouncer.java
SamCarlberg: ... and of course `ScheduledExecutor` doesn't accept `Duration` arguments :disappointed:

app/src/main/java/edu/wpi/first/shuffleboard/app/Shuffleboard.java
```diff
@@ -41,22 +52,53 @@ public void init() throws AlreadyLockedException, IOException {
...
+   * Attempts to loads plugins from all jars found in in the {@link Storage#PLUGINS_DIR plugin directory}. This will
+   * overwrite pre-existing plugins with the same ID string (eg "edu.wpi.first.shuffleboard.Base") in encounter order,
+   * which is alphabetical by jar name. For example, if a jar file "my_plugins.jar" defines a plugin with ID "foo.bar"
+   * and another jar file "more_plugins.jar" <i>also</i> defines a plugin with that ID, the plugin from "more_plugins"
+   * will be loaded first, then unloaded and replaced with the one from "my_plugins.jar". For this reason, plugin
+   * authors should be careful to use unique group IDs. We recommend Java's reverse-DNS naming scheme.
+   *
+   * @throws IOException if the plugin directory could not be read
+   */
+  private void loadPluginsFromDir() throws IOException {
```
nightpool: shouldn't this code be in PluginLoader?

app/src/main/java/edu/wpi/first/shuffleboard/app/plugin/PluginLoader.java
```diff
@@ -0,0 +1,189 @@
...
+  public void load(Plugin plugin) {
+    if (loadedPlugins.contains(plugin)) {
+      throw new IllegalArgumentException("The plugin " + plugin + " is already loaded");
+    }
+    log.info("Loading plugin " + plugin.fullIdString()); //NOPMD log not in if statement
+    plugin.getDataTypes().forEach(DataTypes.getDefault()::register);
+    plugin.getSourceTypes().forEach(SourceTypes.getDefault()::register);
+    plugin.getTypeAdapters().forEach(Serializers::add);
+    plugin.getWidgets().forEach(Widgets.getDefault()::register);
+    plugin.getDefaultWidgets().forEach(Widgets.getDefault()::setDefaultWidget);
```
nightpool: hardcoding all of these seems not great, but i'm not sure what the best way to do fix this is.

api/src/main/java/edu/wpi/first/shuffleboard/api/data/DataTypes.java
```diff
@@ -1,95 +1,225 @@
...
+   *
+   * {@code closestTo(Object.class)} would be sorted as {@code [Bar, Buq], Baz, Foo}. The order of Bar, Baz is not
+   * deterministic and depends on the ordering of the source collection, hence the brackets.
+   *
+   * <p><b>This method does <i>not</i> support comparison of interfaces</b></p>
+   *
+   * @param target the class that should be compared to to determine ordering
+   */
+  @VisibleForTesting
+  static Comparator<Class<?>> closestTo(Class<?> target) {
```
JLLeitschuh: I disagree with that mentality. It makes sense to write tests where you need to make sure something works correctly.

api/src/main/java/edu/wpi/first/shuffleboard/api/data/DataTypes.java
```diff
@@ -1,95 +1,225 @@
...
+      if (naive.isPresent()) {
+        return naive;
+      } else {
+        // Check Java class hierarchy
+        Comparator<Class<?>> classComparator = closestTo(type);
+        List<DataType> sorted = dataTypes.values().stream()
+            .filter(t -> t != All)
+            .sorted((t1, t2) -> classComparator.reversed().compare(t1.getJavaClass(), t2.getJavaClass()))
+            .collect(Collectors.toList());
+        return Optional.of(sorted.get(0));
```
JLLeitschuh: Can't you just call `first()` on the set and get an optional?
Also, should this be `Optional.ofNullable`?

api/src/main/java/edu/wpi/first/shuffleboard/api/util/Registry.java
```diff
@@ -0,0 +1,85 @@
...
+    Objects.requireNonNull(items, "items");
+    for (T item : items) {
```
JLLeitschuh: The array must not be null but you could pass `registerAll(item1, null, item2, null)` what would happen in that case? Should you check each of the args for null too?

api/src/main/java/edu/wpi/first/shuffleboard/api/util/TestUtils.java
```diff
@@ -0,0 +1,30 @@
...
+  private TestUtils() {
+    throw new UnsupportedOperationException("TestUtils is a utility class!");
+  }
+
+  /**
+   * Asserts that this method's call stack originated from a test method invocation.
+   *
+   * @throws IllegalStateException if this method is not called from within a test
+   */
+  public static void assertRunningFromTest() throws IllegalStateException {
```
JLLeitschuh: Yea, wait... What is going on here???

api/src/main/java/edu/wpi/first/shuffleboard/api/widget/Widgets.java
```diff
@@ -78,26 +91,46 @@ public Widget get() {
...
 
   /**
    * Validates a widget description.
    *
    * @param description the description to validate
+   *
    * @throws InvalidWidgetException if the widget is invalid
    */
-  private static void validate(Description description) throws InvalidWidgetException {
+  private void validate(Description description) throws InvalidWidgetException {
```
JLLeitschuh: Why is this no longer static? It doesn't use any member variables.

api/src/main/java/edu/wpi/first/shuffleboard/api/widget/Widgets.java
```diff
@@ -107,40 +140,55 @@ private static void validate(Description description) throws InvalidWidgetExcept
...
   }
 
   /**
+   * Gets a list of the active widgets in the application.
+   *
+   * <p><strong>Do not keep a reference to this list.</strong> It prevents garbage collection of widget instances.</p>
+   */
+  public List<Widget> getActiveWidgets() {
+    // Use a copy; don't want elements in the list to be removed by GC while someone's using it
+    return new ArrayList<>(activeWidgets.keySet());
```
JLLeitschuh: Why not use an `ImmutableList.copyOf`?

api/src/test/java/edu/wpi/first/shuffleboard/api/sources/recording/SerializationTest.java
```diff
@@ -58,51 +59,12 @@ public void testBooleanToBytes() {
...
-
-  @Test
-  public void testStringArrayFromBytes() {
-    byte[] empty = new byte[4];
-    String[] expected = {};
-    assertArrayEquals(expected, Serialization.readStringArray(empty, 0));
-
-    expected = new String[]{"foo", "bar"};
-    assertArrayEquals(expected, Serialization.readStringArray(fooBarBytes, 0));
-  }
```
JLLeitschuh: Why were all these tests removed? Are they now irrelevant?

app/src/main/java/edu/wpi/first/shuffleboard/app/plugin/PluginLoader.java
```diff
@@ -0,0 +1,189 @@
...
+   * Loads a plugin jar and loads all plugin classes within.
+   *
+   * @param jarUri a URI representing  jar file to load plugins from
+   *
+   * @throws IOException if the jar file denoted by the URI could not be found or read
+   */
+  public void loadPluginJar(URI jarUri) throws IOException {
+    URL url = jarUri.toURL();
+    PrivilegedAction<URLClassLoader> getClassLoader = () -> {
+      return new URLClassLoader(new URL[]{url}, ClassLoader.getSystemClassLoader());
```
JLLeitschuh: Is this using the class loader that the rest of the code is using?

I know there are some wacky edge cases you need to be aware of when dealing with class loaders and the garbage collector.

api/src/test/java/edu/wpi/first/shuffleboard/api/sources/recording/SerializationTest.java
SamCarlberg: They were dependent on serializers that were moved to the base plugin

app/src/main/java/edu/wpi/first/shuffleboard/app/plugin/PluginLoader.java
JLLeitschuh: Are you concerned that these are all using the static singleton of the `DataTypes`, `SourceTypes` ect.. objects?

I think the solution is to pas them as dependencies into the object and apply the dependency inversion principle here.

app/src/main/java/edu/wpi/first/shuffleboard/app/plugin/PluginLoader.java
```diff
@@ -0,0 +1,189 @@
...
+          return plugin.getWidgets().contains(w.getClass())
+              || plugin.getDataTypes().contains(source.getDataType())
+              || plugin.getSourceTypes().contains(source.getType());
+        })
+        .filter(w -> SourceTypes.getDefault().isRegistered(w.getSource().getType()))
+        .forEach(w -> tryRestoreSource(w, (DestroyedSource) w.getSource()));
+    plugin.getThemes().forEach(Themes.getDefault()::register);
+
+    plugin.onLoad();
+    plugin.setLoaded(true);
```
JLLeitschuh: Why is this necessary?

app/src/main/java/edu/wpi/first/shuffleboard/app/plugin/PluginLoader.java
SamCarlberg: I'd have to look into that. This is basically the same code that the original dashboard uses for loading jars at runtime and that has worked for the past 6+ years

app/src/main/java/edu/wpi/first/shuffleboard/app/sources/DestroyedSource.java
```diff
@@ -0,0 +1,110 @@
...
+import javafx.beans.property.ObjectProperty;
+import javafx.beans.property.Property;
+import javafx.beans.property.SimpleBooleanProperty;
+import javafx.beans.property.SimpleObjectProperty;
+import javafx.beans.property.SimpleStringProperty;
+import javafx.beans.property.StringProperty;
+
+/**
+ * A type of data source that represents the state of another source that has been destroyed or removed as a result
+ * of its defining plugin being unloaded. The data source may
```
JLLeitschuh: Incomplete javadoc.

app/src/test/java/edu/wpi/first/shuffleboard/app/json/DashboardTabPaneSaverTest.java
```diff
@@ -19,11 +19,12 @@
 
 import static org.junit.jupiter.api.Assertions.assertEquals;
 
+@Disabled("Depends on network tables")
```
JLLeitschuh: Do you have a plan to re-enable this?

app/src/test/java/edu/wpi/first/shuffleboard/app/plugin/PluginLoaderTest.java
```diff
@@ -0,0 +1,30 @@
...
+    public MockPlugin() {
+      super("test", "Mock Plugin", "0.0.0", "A plugin for testing");
+    }
+
+  }
+
+  @Test
+  public void testLoadClass() {
+    PluginLoader loader = new PluginLoader();
+    boolean loaded = loader.loadPluginClass(MockPlugin.class);
```
JLLeitschuh: As soon as you try doing this with more complicated classes that actually have plugin logic you are going to end up adding stuff to the `Widgets` singleton and other singletons which will affect other tests.

plugins/base/src/test/java/edu/wpi/first/shuffleboard/plugin/base/widget/AbstractWidgetTest.java
```diff
@@ -0,0 +1,32 @@
...
+    }
+
+    Widgets.setDefault(widgetRegistry);
+    DataTypes.setDefault(dataTypeRegistry);
+  }
+
+  @AfterAll
+  public static void resetRegistries() {
+    Widgets.setDefault(new Widgets());
+    DataTypes.setDefault(new DataTypes());
```
JLLeitschuh: This is where static singletons bite you in the butt. You really shouldn't need to do this.

plugins/networktables/networktables.gradle.kts
JLLeitschuh: This has been resolved.

api/src/test/java/edu/wpi/first/shuffleboard/api/sources/recording/SerializationTest.java
JLLeitschuh: Okay.

api/src/main/java/edu/wpi/first/shuffleboard/api/util/Debouncer.java
```diff
@@ -0,0 +1,61 @@
...
+import java.util.concurrent.ScheduledExecutorService;
+import java.util.concurrent.ScheduledFuture;
+import java.util.concurrent.TimeUnit;
+
+/**
+ * A simple class for debouncing method calls. Debouncing is used to prevent an expensive method from being called
+ * in rapid succession, only allowing it to run after a certain amount of time has passed without it being called
+ * again.
+ */
+public class Debouncer {
```
SamCarlberg: I don't think we need to add a redundant interface

app/src/main/java/edu/wpi/first/shuffleboard/app/MainWindowController.java
```diff
@@ -138,6 +127,121 @@ private void initialize() throws IOException {
...
+
+    // Add widgets to the gallery as well
+    widgetGallery.setWidgets(Widgets.getDefault().allWidgets());
+  }
+
+  /**
+   * Removes all traces from a plugin from the application window. Source trees will be removed and all widgets
+   * defined by the plugin will be removed from all dashboard tabs.
+   */
+  private void tearDown(Plugin plugin) {
```
SamCarlberg: Bind widgetGallery how? We'd have to pass it into the plugin loader when the controller is initialized

app/src/main/java/edu/wpi/first/shuffleboard/app/plugin/PluginLoader.java
SamCarlberg: As I keep saying, I want to move these all to use a DI framework like Guice. We don't have time for that right now

api/src/main/java/edu/wpi/first/shuffleboard/api/data/DataTypes.java
SamCarlberg: Ah, that may work. `ofNullable` is pointless since elements should never be null

api/src/main/java/edu/wpi/first/shuffleboard/api/util/Registry.java
SamCarlberg: `register` does a null check....

api/src/main/java/edu/wpi/first/shuffleboard/api/plugin/Plugin.java
```diff
@@ -0,0 +1,155 @@
...
+    this.name = name;
+    this.version = version;
+    this.description = description;
+  }
+
+  /**
+   * The group ID of this plugin. For example, the stock plugins have the group ID {@code edu.wpi.first.shuffleboard}.
+   * This must combine with {@link #getName()} to create a unique identifier for this plugin.
+   */
+  public final String getGroupId() {
```
SamCarlberg: DataTypes absolutely cannot, because those names are used in the LiveWindow to say the data type (eg `Analog Input`, `Speed Controller`, etc). Allowing duplicate names would at best use the first registered type or at worst use the most recently registered type.

The widgets issue is a UX one. Seeing multiple instances of the same widget name in a list is confusing. It's impossible to know which name is the one you want. Users may also think duplicate names is a bug and not expect to get different widgets for different names. We have an option of displaying them like plugins, with the name preceded by a group ID. But that's a terrible UX:

```
Show as...
  edu.wpi.first.shuffleboard.BooleanBox
  edu.wpi.first.shuffleboard.ToggleSwitch
  third.party.plugin.BooleanBox
  third.party.plugin.ToggleSwitch
  ToggleButton
```

api/src/main/java/edu/wpi/first/shuffleboard/api/data/DataType.java
```diff
@@ -9,52 +11,70 @@
  *
  * @see DataTypes
  */
-public interface DataType<T> {
+public abstract class DataType<T> {
```
SamCarlberg: What other things? I don't agree that multiple inheritance is a usecase we should worry about for data types

api/src/main/java/edu/wpi/first/shuffleboard/api/util/Debouncer.java
SamCarlberg: I like that. I'll still call it `Debouncer` though, because it's still fundamentally debouncing a supplied function call

app/src/main/java/edu/wpi/first/shuffleboard/app/plugin/PluginLoader.java
JLLeitschuh: You can apply the dependency inversion principle without using a DI framework.

api/src/main/java/edu/wpi/first/shuffleboard/api/util/Registry.java
JLLeitschuh: Okay! Cool!

api/src/main/java/edu/wpi/first/shuffleboard/api/util/Debouncer.java
JLLeitschuh: You should still accept it as the argument and split it out if you need to.

api/src/main/java/edu/wpi/first/shuffleboard/api/util/TestUtils.java
SamCarlberg: Basically just used to make sure the default registries aren't changed during runtime. We really need to get DI set up

app/src/test/java/edu/wpi/first/shuffleboard/app/json/DashboardTabPaneSaverTest.java
SamCarlberg: Yeah. I'll need to change the save file to use mock widgets

api/src/main/java/edu/wpi/first/shuffleboard/api/util/TestUtils.java
JLLeitschuh: I searched your changes and this method isn't called anywhere.

If this is really what you wanted wouldn't you do something like `assertNotRunningFromTest`? Even that sounds really bad and should be avoided.

You can apply the dependency inversion principle without using a DI framework. At least to start you can begin by creating constructors that take the objects they depend upon and still have a default no-arg constructor that uses the scope singletons for the time being.

Also, this method won't work when you have a method called on the javaFX thread in a unit test.

app/src/test/java/edu/wpi/first/shuffleboard/app/json/WidgetSaverTest.java
AustinShalit: https://github.com/TestFX/TestFX/pull/412#issuecomment-323614936

PR #154, opened 2017-08-22
Remove tiles by dragging outside of pane

### LINKED ISSUE #134: Dragging a tile off the grid should remove it

SamCarlberg:
[Video](https://www.youtube.com/watch?v=tayKhgzKX8Q)

Closes #134

### DIFF
--- a/app/src/main/java/edu/wpi/first/shuffleboard/app/WidgetPaneController.java
+++ b/app/src/main/java/edu/wpi/first/shuffleboard/app/WidgetPaneController.java
@@ -1,8 +1,5 @@
 package edu.wpi.first.shuffleboard.app;
 
-import edu.wpi.first.shuffleboard.app.components.TileLayout;
-import edu.wpi.first.shuffleboard.app.components.WidgetPane;
-import edu.wpi.first.shuffleboard.app.components.WidgetTile;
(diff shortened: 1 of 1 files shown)


app/src/main/java/edu/wpi/first/shuffleboard/app/WidgetPaneController.java
```diff
@@ -124,6 +124,23 @@ private void initialize() {
       event.consume();
     });
 
+    // Drag tiles outside the pane to remove them
+    pane.setOnDragExited(e -> {
```
nightpool: I'm worried that this will remove a widget as soon as we drag it away, instead of removing a widget when we *release* the drag.

SamCarlberg: The widget pane DnD handler isn't able to know when the drag gesture finishes if it's not on the widget pane. It only knows when the drag exits, enters, or is released _on the widget pane_.

nightpool:
Definitely interested in this feature, tracked by #134, but this implementation has too many foot guns to review it right now. 

Consensus was something like serializing the tile to JSON and removing it from the widget pane during the drag gesture is probably the correct move.

PR #156, opened 2017-08-23
Add PDP widget

SamCarlberg:
This is a big widget (doesn't really work with tile size < ~350x250). Semi-depends on #155 to make sure the tile doesn't get smaller than the widget

## Screenshots

pdp widget screenshot


plugins/base/src/main/java/edu/wpi/first/shuffleboard/plugin/base/widget/PowerDistributionPanelWidget.java
```diff
@@ -0,0 +1,131 @@
...
+  private LinearIndicator voltage;
+  @FXML
+  private LinearIndicator totalCurrent;
+  @FXML
+  private Label voltageText;
+  @FXML
+  private Label totalCurrentText;
+
+  // Arrange the channel indicators in an array for ease of use
+  private LinearIndicator[] channels;
```
JLLeitschuh: `ImmutableList`?

JLLeitschuh:
Why is this in the base plugin instead of in the network tables plugin? Or in its own plugin?

SamCarlberg:
All builtin widgets are in the base plugin, with the exception of `NetworkTable Tree` because that is specifically for NetworkTables

JLLeitschuh:
Where does the PDP data get loaded from? Is that code missing from the PR?

SamCarlberg:
It's provided by whatever source. The whole point of having the source API is to remove the burden of knowing where the data comes from and how it's loaded and created.

JLLeitschuh:
1. What if I want to write a plugin that creates custom data?
 2. I understand that but where will you get the PDP data from when this runs live? A Live Window plugin?

SamCarlberg:
1. Then you create a custom data type and a custom data object for that type
2. From any source, usually network tables

JLLeitschuh:
Are you expecting users to publish PDP data in a specific way to get this to work? How is a user supposed to tie shuffleboard to the published PDP data so that this works correctly. This may be my basic lack of understanding the underlying architecture.

nightpool:
@JLLeitschuh You're correct that eventually we should document the expected maps for datatypes, but as is currently we're only using data types that are already expressly supported by SmartDashboard so I don't think it's a very high priority.

PR #157, opened 2017-08-24
Change recording file format to {date}/{time}.frc

### LINKED ISSUE #50: Recording files can't be saved on Windows
Due to the file name format `dashboard_recording_uuuu-MM-dd_HH:mm:ss` and Windows not allowing colons in file names

SamCarlberg:
Fix #50
Remove unused serialization methods

Recording files are now saved to `~/SmartDashboard/recordings/<date>/recording-<time>.frc`, where
`date`  is formatted in `YYYY-MM-DD` (eg `2017-01-01`) and `time` is formatted in `HH.mm.ss` (eg `12.34.56` for 12:34:56). Periods are used instead of colons because Windows doesn't allow colons in file names

Sample file structure:

```
SmartDashboard
  recordings
    2017-08-24
      recording-12.15.20.frc
      recording-13.41.51.frc
    2017-08-25
      recording-09.31.45.frc
      recording-10.12.33.frc
```

### DIFF
--- a/api/src/main/java/edu/wpi/first/shuffleboard/api/sources/recording/Recorder.java
+++ b/api/src/main/java/edu/wpi/first/shuffleboard/api/sources/recording/Recorder.java
@@ -26,8 +26,9 @@ public final class Recorder {
 
   private static final Logger log = Logger.getLogger(Recorder.class.getName());
 
-  private static final DateTimeFormatter timeFormatter =
-      DateTimeFormatter.ofPattern("uuuu-MM-dd_HH:mm:ss", Locale.getDefault());
+  private static final DateTimeFormatter dateFormatter = DateTimeFormatter.ISO_DATE;
+  private static final DateTimeFormatter timeFormatter = DateTimeFormatter.ofPattern("HH.mm.ss", Locale.getDefault());
+
   private static final Recorder instance = new Recorder();
 
   private final BooleanProperty running = new SimpleBooleanProperty(this, "running", false);
@@ -54,7 +55,9 @@ public final class Recorder {
       return;
     }
     try {
-      String file = String.format(Storage.RECORDING_FILE_FORMAT, createTimestamp());
+      String date = dateFormatter.format(LocalDateTime.ofInstant(startTime, ZoneId.systemDefault()));
+      String time = timeFormatter.format(LocalDateTime.ofInstant(startTime, ZoneId.systemDefault()));
+      String file = Storage.RECORDING_FILE_FORMAT.replace("${date}", date).replace("${time}", time);
       Serialization.saveRecording(recording, file);
       log.fine("Saved recording to " + file);
     } catch (IOException e) {
--- a/api/src/main/java/edu/wpi/first/shuffleboard/api/sources/recording/Recorder.java
+++ b/api/src/main/java/edu/wpi/first/shuffleboard/api/sources/recording/Recorder.java
@@ -7,10 +7,6 @@ import edu.wpi.first.shuffleboard.api.util.ThreadUtils;
 
 import java.io.IOException;
 import java.time.Instant;
-import java.time.LocalDateTime;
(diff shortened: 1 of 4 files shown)


SamCarlberg:
@JLLeitschuh What the heck is FindBugs doing here?

```xml
<?xml version="1.0" encoding="UTF-8"?><BugInstance type="NP_NULL_ON_SOME_PATH_FROM_RETURN_VALUE" priority="2" rank="13" abbrev="NP" category="STYLE">
  <Class classname="edu.wpi.first.shuffleboard.api.sources.recording.Serialization">
    <SourceLine classname="edu.wpi.first.shuffleboard.api.sources.recording.Serialization" start="27" end="408" sourcefile="Serialization.java" sourcepath="edu/wpi/first/shuffleboard/api/sources/recording/Serialization.java"/>
  </Class>
  <Method classname="edu.wpi.first.shuffleboard.api.sources.recording.Serialization" name="saveRecording" signature="(Ledu/wpi/first/shuffleboard/api/sources/recording/Recording;Ljava/lang/String;)V" isStatic="true">
    <SourceLine classname="edu.wpi.first.shuffleboard.api.sources.recording.Serialization" start="51" end="88" startBytecode="0" endBytecode="873" sourcefile="Serialization.java" sourcepath="edu/wpi/first/shuffleboard/api/sources/recording/Serialization.java"/>
  </Method>
  <LocalVariable name="?" register="-1" pc="373" role="LOCAL_VARIABLE_UNKNOWN"/>
  <SourceLine classname="edu.wpi.first.shuffleboard.api.sources.recording.Serialization" start="86" end="86" startBytecode="373" endBytecode="373" sourcefile="Serialization.java" sourcepath="edu/wpi/first/shuffleboard/api/sources/recording/Serialization.java" role="SOURCE_LINE_INVOKED"/>
  <SourceLine classname="edu.wpi.first.shuffleboard.api.sources.recording.Serialization" start="86" end="86" startBytecode="364" endBytecode="364" sourcefile="Serialization.java" sourcepath="edu/wpi/first/shuffleboard/api/sources/recording/Serialization.java" role="SOURCE_LINE_KNOWN_NULL"/>
</BugInstance>
```

[The relevant line](https://github.com/wpilibsuite/shuffleboard/pull/157/files#diff-68f7e72addbf1762acba66d480f79d8eR86)

JLLeitschuh:
So [ISO standard](https://en.wikipedia.org/wiki/ISO_8601) but with dots instead of colons?

JLLeitschuh:
I think `Paths.get(file).getParent()` something might return null and you don't check for null?

SamCarlberg:
> So ISO standard but with dots instead of colons?

Exactly

> I think Paths.get(file).getParent() something might return null and you don't check for null?

Huh, looks like it does (at least for Windows). Weird how that's not specified as part of the API.

> I do not think this is a good idea. It makes files much harder to search for.

How? I've found that putting every file in the same directory gets very cluttered very fast without manually organizing or deleting the recording files.

api/src/main/java/edu/wpi/first/shuffleboard/api/sources/recording/Recorder.java
```diff
@@ -54,7 +55,9 @@ private void saveToDisk() {
       return;
     }
     try {
-      String file = String.format(Storage.RECORDING_FILE_FORMAT, createTimestamp());
+      String date = dateFormatter.format(LocalDateTime.ofInstant(startTime, ZoneId.systemDefault()));
+      String time = timeFormatter.format(LocalDateTime.ofInstant(startTime, ZoneId.systemDefault()));
+      String file = Storage.RECORDING_FILE_FORMAT.replace("${date}", date).replace("${time}", time);
```
nightpool: I would prefer if this code (and the DateTimeFormatters) were in the Storage class, so that it has an api like File getRecordingFile(startTime)

PR #164, opened 2017-08-27
List Layout

nightpool:
Still left to do (either here or in a followup)

 - [ ] more types of layouts (probably at least an hbox--I couldn't think of a good name)
 - [ ] ways of removing/reordering widgets in a layout

I'm planning on implementing these tonight (at least some of B, and probably A)


app/src/main/java/edu/wpi/first/shuffleboard/app/components/WidgetPane.java
```diff
@@ -105,8 +107,10 @@ public WidgetPane() {
    *
    * @param predicate the predicate to use to find the desired widget tile
    */
-  public Optional<WidgetTile> tileMatching(Predicate<WidgetTile> predicate) {
+  public Optional<Tile> tileMatching(Predicate<Tile> predicate) {
     return tiles.stream()
+                .map(TypeUtils.optionalCast(Tile.class))
+                .filter(Optional::isPresent).map(Optional::get)
```
SamCarlberg: May also want to add a `TypeUtils.optionalStream` method to flatmap the returned `Optional` as a stream like:

```java
.flatMap(TypeUtils::optionalStream)
```

app/src/main/java/edu/wpi/first/shuffleboard/app/components/WidgetPane.java
```diff
@@ -125,7 +129,8 @@ private void resizeTiles() {
    * Gets the tile for the widget containing the given source.
    */
   public Optional<WidgetTile> widgetForSource(DataSource<?> source) {
```
SamCarlberg: I think this should return all tiles, not just widget tiles, that contain a widget for the source.

app/src/main/java/edu/wpi/first/shuffleboard/app/components/WidgetPane.java
```diff
@@ -153,14 +158,10 @@ public WidgetTile addWidget(Widget widget, TileSize size) {
   }
 
   /**
-   * Adds a widget to the tile view in the specified location. The tile will be the specified
-   * size.
-   *
-   * @param widget the widget to add
-   * @param size   the size of the tile used to display the widget
+   * Add an arbitrary component to the WidgetPane.
```
SamCarlberg: Why was documentation removed?

app/src/main/java/edu/wpi/first/shuffleboard/app/components/WidgetPane.java
```diff
@@ -193,14 +194,20 @@ public TileSize sizeOfWidget(Widget widget) {
             (int) (height / getTileSize()));
   }
 
-  public void removeWidget(WidgetTile tile) {
+  /**
+   * Remove a given tile from the widget pane.
```
SamCarlberg: This should have a `@return` tag because what this returns is not obvious without looking at the source

app/src/main/resources/edu/wpi/first/shuffleboard/app/components/Tile.fxml
```diff
@@ -0,0 +1,16 @@
...
+<?import edu.wpi.first.shuffleboard.app.components.Tile?>
+<?import javafx.scene.layout.StackPane?>
+
+<fx:root xmlns="http://javafx.com/javafx"
+         xmlns:fx="http://javafx.com/fxml"
+         type="Tile"
+         fx:id="tile">
+    <top>
+        <StackPane styleClass="tile-title-bar">
+            <EditableLabel fx:id="titleLabel" styleClass="tile-title-label"/>
```
SamCarlberg: 👍 for editability

plugins/base/src/main/java/edu/wpi/first/shuffleboard/plugin/base/layout/ListLayout.java
```diff
@@ -0,0 +1,80 @@
...
+  private Subscription retained;
+
+  @FXML
+  private void initialize() {
+    retained = EasyBind.listBind(container.getChildren(), EasyBind.map(widgets, this::paneFor));
+  }
+
+  private Pane paneFor(Component widget) {
+    BorderPane pane = new BorderPane(widget.getView());
+    pane.getStyleClass().add("layout--stack");
```
SamCarlberg: Why the double dash? That's not normal syntax

api/src/main/java/edu/wpi/first/shuffleboard/api/widget/Widgets.java
```diff
@@ -29,13 +31,13 @@
 /**
  * Utility class for keeping track of known widgets.
  */
-public class Widgets extends Registry<Class<? extends Widget>> {
+public class Widgets extends Registry<ComponentType> {
```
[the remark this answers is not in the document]
SamCarlberg: This class should be renamed to `Components` since it's no longer responsible for only widgets

api/src/main/java/edu/wpi/first/shuffleboard/api/widget/Component.java
```diff
@@ -0,0 +1,25 @@
+package edu.wpi.first.shuffleboard.api.widget;
+
+import java.util.stream.Stream;
+
+import javafx.beans.property.Property;
+import javafx.scene.layout.Pane;
+
+public interface Component {
+  Pane getView();
```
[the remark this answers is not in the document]
nightpool: this was taken unchanged from the Widget API.

api/src/main/java/edu/wpi/first/shuffleboard/api/widget/LayoutClass.java
```diff
@@ -0,0 +1,32 @@
+package edu.wpi.first.shuffleboard.api.widget;
+
+import java.util.logging.Level;
+import java.util.logging.Logger;
+
+public class LayoutClass<T extends Layout> implements ComponentType {
```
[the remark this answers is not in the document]
nightpool: renamed to avoid confusion.

api/src/main/java/edu/wpi/first/shuffleboard/api/widget/Component.java
```diff
@@ -0,0 +1,25 @@
+package edu.wpi.first.shuffleboard.api.widget;
+
+import java.util.stream.Stream;
+
+import javafx.beans.property.Property;
+import javafx.scene.layout.Pane;
+
+public interface Component {
```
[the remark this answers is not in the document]
JLLeitschuh: Javadoc would be very useful

api/src/main/resources/edu/wpi/first/shuffleboard/api/base.css
```diff
@@ -83,19 +83,33 @@
...
-    -fx-font-size: 12pt;
+    -fx-font-size: 11pt;
 }
 
 .tile .property-sheet * {
     /* TODO figure out exactly what needs to be set */
     -fx-background-color: transparent;
 }
 
+/********/
```
JLLeitschuh: Why the line of stars in the comment?

app/src/main/java/edu/wpi/first/shuffleboard/app/components/Tile.java
```diff
@@ -0,0 +1,109 @@
...
+  private final BooleanProperty selected = new PseudoClassProperty(this, "selected");
+
+  /**
+   * Creates an empty tile. The content and size must be set with {@link #setContent(T)} and
+   * {@link #setSize(TileSize)}.
+   */
+  protected Tile() {
+    try {
+      FXMLLoader loader = new FXMLLoader(Tile.class.getResource("Tile.fxml"));
+      loader.setRoot(this);
```
JLLeitschuh: Is this guaranteed to run in the JavaFX thread?

Also, you are exposing a reference to `this` within the constructor which is bad practice as it can lead to race conditions regarding fields that should be final (and thus non-null) being null when viewed on another thread.

We had this problem with GRIP:
https://github.com/WPIRoboticsProjects/GRIP/issues/155


[Guava Google Groups Discussion with Dev](https://groups.google.com/forum/#!topic/guava-discuss/4ZJhhpmMvKQ)

> ## Don't publish the "this" reference during construction
> 
> One of the mistakes that can introduce a data race into your class is to expose the this reference to another thread before the constructor has completed. Sometimes the reference is explicit, such as directly storing this in a static field or collection, but other times it can be implicit, such as when you publish a reference to an instance of a non-static inner class in a constructor. Constructors are not ordinary methods -- they have special semantics for initialization safety. An object is assumed to be in a predictable, consistent state after the constructor has completed, and publishing a reference to an incompletely constructed object is dangerous. Listing 2 shows an example of introducing this sort of race condition into a constructor. It may look harmless, but it contains the seeds of serious concurrency problems.
> -[Brian Goetz](http://www.ibm.com/developerworks/library/j-jtp0618/)

app/src/main/java/edu/wpi/first/shuffleboard/app/components/WidgetPane.java
JLLeitschuh: I agree! This seems to be a common pattern that this code base utilizes a lot. Turning it into a utility makes a lot of sense.

app/src/main/java/edu/wpi/first/shuffleboard/app/json/LayoutSaver.java
```diff
@@ -0,0 +1,52 @@
...
+
+  @Override
+  public Layout deserialize(JsonElement json, JsonDeserializationContext context) throws JsonParseException {
+    String name = json.getAsJsonObject().get("_type").getAsString();
+    JsonArray children = json.getAsJsonObject().get("_children").getAsJsonArray();
+
+    Layout layout = Widgets.getDefault().createComponent(name).flatMap(TypeUtils.optionalCast(Layout.class))
+        .orElseThrow(() -> new JsonParseException("Can't find layout name " + name));
+
+    children.forEach(o -> {
```
JLLeitschuh: Something more descriptive than `o` here would be useful.

app/src/main/java/edu/wpi/first/shuffleboard/app/components/Tile.java
nightpool: this is the correct and documented way to construct JavaFX custom components. see https://docs.oracle.com/javase/8/javafx/fxml-tutorial/custom_control.htm

api/src/main/java/edu/wpi/first/shuffleboard/api/widget/Widgets.java
```diff
@@ -58,80 +60,80 @@ public static void setDefault(Widgets instance) {
...
+      throw new IllegalArgumentException("Component class " + type.getClass().getName() + " is already registered");
     }
+    widgets.put(type.getName(), type);
+    addItem(type);
+  }
+
+  /**
+   * Convenience overload for registering annotated widgets.
+   */
+  public <T extends Widget> void register(Class<T> widgetClass) {
```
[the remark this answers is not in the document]
nightpool: It would streamline this API at the cost of making the Plugins API *much* harder to work with. I'm not sure it's worth it (especially since you'd also have to do something about `registerDefaultWidget`, and that case seems much less obvious....)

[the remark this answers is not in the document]
nightpool: Not sure what the point of requiring AnnotatedWidget would be either, since it's not required to use the `@Description` annotation—it's just a helper.

app/src/main/java/edu/wpi/first/shuffleboard/app/components/WidgetPane.java
nightpool: unused because it's no longer used. you might have been thinking of dashboard.selectWidgets ?

app/src/main/java/edu/wpi/first/shuffleboard/app/components/WidgetPane.java
nightpool: restored.

api/src/main/java/edu/wpi/first/shuffleboard/api/plugin/Plugin.java
```diff
@@ -109,13 +110,20 @@ public void onUnload() {
   }
 
   /**
-   * Gets a list of the widget types that this plugin defines.
+   * Gets a list of the annotated widget classes that this plugin defines.
```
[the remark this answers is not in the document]
nightpool: see below. I'm not sure that this is the right API change to make from a usability perspective. I would like to keep the "shortcut" way of adding Widgets and then add the more verbose API for plugins that need the flexibility.

app/src/main/java/edu/wpi/first/shuffleboard/app/components/WidgetPane.java
nightpool: fixed.

nightpool:
Review comments addressed except for the ones about the widget API for annotated widgets, which I don't think we should change for plugin usability purposes.

api/src/main/java/edu/wpi/first/shuffleboard/api/components/EditableLabel.java
```diff
@@ -12,12 +14,17 @@
   private final PseudoClassProperty editing = new PseudoClassProperty(this, "editing");
   private final StringProperty text = new SimpleStringProperty(this, "text");
 
+
+
```
AustinShalit: Surprised checkstyle did not find this.  There are too many empty lines here.  One will do.

api/src/main/java/edu/wpi/first/shuffleboard/api/util/TypeUtils.java
```diff
@@ -24,4 +25,37 @@ private TypeUtils() {
...
+   *     .forEach(w -> w.setSource(...))
+   * </code>
+   *
+   * <p>can be turned into:</p>
+   *
+   * <code>
+   *   getComponents()
+   *     .flatMap(TypeUtils.castStream(Widget.class))
+   *     .forEach(w -> w.setSource(...))
+   * </code>
```
AustinShalit: Use: 
```
* <pre>
* {@code
* System.out.println("");
* }
* </pre>
```

api/src/main/java/edu/wpi/first/shuffleboard/api/util/TypeUtils.java
```diff
@@ -24,4 +25,37 @@ private TypeUtils() {
...
+  /**
+   * Allows you to filter out members of a subtype from a stream of some base type.
+   * For example, this code:
+   *
+   * <code>
+   *   getComponents().stream()
+   *     .filter(c -> c instanceof Widget)
+   *     .map(c -> (Widget) c)
+   *     .forEach(w -> w.setSource(...))
+   * </code>
```
AustinShalit: Use: 
```
* <pre>
* {@code
* System.out.println("");
* }
* </pre>
```

api/src/main/java/edu/wpi/first/shuffleboard/api/util/TypeUtils.java
```diff
@@ -24,4 +25,37 @@ private TypeUtils() {
...
+   *
+   * <p>can be turned into:</p>
```
AustinShalit: You do not need the closing tag.

api/src/main/java/edu/wpi/first/shuffleboard/api/util/TypeUtils.java
```diff
@@ -24,4 +25,37 @@ private TypeUtils() {
   public static <T> Function<Object, Optional<T>> optionalCast(Class<T> cls) {
     return value -> optionalCast(value, cls);
   }
+
+  /**
+   * Allows you to filter out members of a subtype from a stream of some base type.
+   * For example, this code:
```
AustinShalit: Line break between description and example.

api/src/main/java/edu/wpi/first/shuffleboard/api/util/TypeUtils.java
```diff
@@ -24,4 +25,37 @@ private TypeUtils() {
   public static <T> Function<Object, Optional<T>> optionalCast(Class<T> cls) {
     return value -> optionalCast(value, cls);
   }
+
+  /**
+   * Allows you to filter out members of a subtype from a stream of some base type.
```
AustinShalit: Usually if documentation uses the word 'you' it could be improved open to make the documentation clearer.

api/src/main/java/edu/wpi/first/shuffleboard/api/widget/ComponentType.java
```diff
@@ -0,0 +1,17 @@
+package edu.wpi.first.shuffleboard.api.widget;
+
+import java.util.function.Supplier;
+
+/**
+ * A ComponentType is a Java object that encapsulates a single "variety" of component that the
```
AustinShalit: Why do programmers need to know that it is a Java object?

api/src/main/java/edu/wpi/first/shuffleboard/api/widget/Components.java
```diff
@@ -52,86 +54,90 @@ public static Widgets getDefault() {
...
-          if (fxml) {
-            FXMLLoader loader = new FXMLLoader(widgetClass.getResource(controller.value()));
-            loader.load();
-            return loader.getController();
-          } else {
+        return viewFor(widgetClass).orElseGet(() -> {
+          try {
             return widgetClass.newInstance();
+          } catch (InstantiationException | IllegalAccessException e) {
+            Logger.getLogger("Components").log(Level.WARNING, "error creating widget", e);
```
AustinShalit: Use the class name here instead of the magic-String.

api/src/main/java/edu/wpi/first/shuffleboard/api/widget/Components.java
```diff
@@ -269,4 +288,26 @@ public void setDefaultWidget(DataType dataType, String widgetName) {
...
+  public static <T> Optional<T> viewFor(Class<T> annotatedClass) {
+    ParametrizedController controller = annotatedClass.getAnnotation(ParametrizedController.class);
+
+    if (controller != null) { //NOPMD readability
+      try {
+        FXMLLoader loader = new FXMLLoader(annotatedClass.getResource(controller.value()));
+        loader.load();
+        return Optional.of(loader.getController());
+      } catch (IOException e) {
+        Logger.getLogger("Components").log(Level.WARNING, "error creating parametrized controller", e);
```
AustinShalit: Use class name instead of magic-string.

api/src/main/java/edu/wpi/first/shuffleboard/api/widget/Components.java
```diff
@@ -269,4 +288,26 @@ public void setDefaultWidget(DataType dataType, String widgetName) {
...
+      }
+    } else {
+      Logger.getLogger("Components").log(Level.WARNING,
```
AustinShalit: Use class name instead of magic-string.

app/src/main/java/edu/wpi/first/shuffleboard/app/WidgetPaneController.java
```diff
@@ -39,15 +47,19 @@
 import javafx.scene.layout.GridPane;
 import javafx.scene.layout.Region;
 
+// needs refactoring to split out per-widget interaction
+@SuppressWarnings("PMD.GodClass")
```
AustinShalit: PMD is right, this class is doing too much right now.

app/src/main/java/edu/wpi/first/shuffleboard/app/components/LayoutTile.java
```diff
@@ -0,0 +1,16 @@
+package edu.wpi.first.shuffleboard.app.components;
+
+import edu.wpi.first.shuffleboard.api.widget.Layout;
+import edu.wpi.first.shuffleboard.api.widget.TileSize;
+
+public class LayoutTile extends Tile<Layout> {
```
AustinShalit: JavaDoc?

app/src/main/java/edu/wpi/first/shuffleboard/app/components/Tile.java
```diff
@@ -0,0 +1,109 @@
...
+
+import java.io.IOException;
+
+import javafx.beans.property.BooleanProperty;
+import javafx.beans.property.Property;
+import javafx.beans.property.SimpleObjectProperty;
+import javafx.fxml.FXMLLoader;
+import javafx.scene.layout.BorderPane;
+
+public class Tile<T extends Component> extends BorderPane {
```
AustinShalit: JavaDoc?

app/src/main/java/edu/wpi/first/shuffleboard/app/components/Tile.java
AustinShalit: Do you want to set the controller to this class also?

api/src/main/java/edu/wpi/first/shuffleboard/api/util/TypeUtils.java
JLLeitschuh: You need a `<p/>` for the javadoc parser to actually add it in the html.

api/src/main/java/edu/wpi/first/shuffleboard/api/util/TypeUtils.java
SamCarlberg: We typically use HTML `<code>` tags for multi-line code blocks like this

AustinShalit: Those do not render correctly in JavaDoc.

api/src/main/java/edu/wpi/first/shuffleboard/api/widget/Widgets.java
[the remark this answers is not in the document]
SamCarlberg: `AnnotatedWidget` requires a `@Description` annotation. Subclasses are out of spec if they don't have that annotation. `@ParamatrizedController` is optional. It's possible to implement `Widget` and use a `@Description` annotation, but if it's to be used they'll be reimplementing `AnnotatedWidget` anyway.

This method will also throw an exception if the class _isn't_ annotated with `@Description`, so it's better for compile-time safety to limit this to `Class<? extends AnnotatedWidget>`

app/src/main/java/edu/wpi/first/shuffleboard/app/components/Tile.java
```diff
@@ -0,0 +1,109 @@
...
+  }
+
+  public BooleanProperty selectedProperty() {
+    return selected;
+  }
+
+  /**
+   * Create a tile for an arbitrary component.
+   */
+  public static Tile<?> tileFor(Component component, TileSize size) {
```
SamCarlberg: It may be beneficial to use a bounded wildcard here so consumers can use

```java
Tile<Widget> widgetTile = Tile.tileFor(someWidget, size);
```

api/src/main/java/edu/wpi/first/shuffleboard/api/widget/Widgets.java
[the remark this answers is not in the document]
nightpool: > It's possible to implement Widget and use a @Description annotation, but if it's to be used they'll be reimplementing AnnotatedWidget anyway.

there are reasons people would want to do that (multiple inheritance, for example)

if we don't need the actual type (and we don't) why require it?

api/src/main/java/edu/wpi/first/shuffleboard/api/widget/ComponentType.java
nightpool: Because it's distinct from a Java Type, which is what you'd normally expect this to be

api/src/main/java/edu/wpi/first/shuffleboard/api/widget/Components.java
nightpool: None of our other code does anything like this. Consistency is important.

app/src/main/java/edu/wpi/first/shuffleboard/app/WidgetPaneController.java
nightpool: Yes but I can't address that in this PR.

app/src/main/java/edu/wpi/first/shuffleboard/app/components/Tile.java
nightpool: no???

nightpool: No class should ever be both a component and a controller.

api/src/main/java/edu/wpi/first/shuffleboard/api/widget/LayoutType.java
```diff
@@ -0,0 +1,12 @@
+package edu.wpi.first.shuffleboard.api.widget;
+
+/**
+ * A typed sub-class of ComponentType for Layouts. This interface serves as a runtime
+ * "marker type" for types that produce layouts.
+ *
+ * <p>Inheriting from this interface is required for a Component to show up in the
+ * "Add to new layout..." menu, among other things.
+ */
+public interface LayoutType extends ComponentType {
```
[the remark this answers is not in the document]
nightpool: Interfaces and classes are completely different things.... LayoutClass is a concrete implementation of the LayoutType interface that requires the Layout to be able to be instantiated by reflection.

app/src/main/java/edu/wpi/first/shuffleboard/app/components/Tile.java
nightpool: will fix.

api/src/main/java/edu/wpi/first/shuffleboard/api/widget/Components.java
AustinShalit: Examples please?
https://github.com/wpilibsuite/shuffleboard/blob/1197c65e2e6a283122ab73f5524e1da5b8b77708/plugins/base/src/main/java/edu/wpi/first/shuffleboard/plugin/base/widget/VoltageViewWidget.java#L26
https://github.com/wpilibsuite/shuffleboard/blob/b2841e0686dd858be2bbfdddfa123095134f8f9f/api/src/main/java/edu/wpi/first/shuffleboard/api/sources/recording/Recorder.java#L23
https://github.com/wpilibsuite/shuffleboard/blob/1197c65e2e6a283122ab73f5524e1da5b8b77708/app/src/main/java/edu/wpi/first/shuffleboard/app/PluginPaneController.java#L34
https://github.com/wpilibsuite/shuffleboard/blob/1197c65e2e6a283122ab73f5524e1da5b8b77708/app/src/main/java/edu/wpi/first/shuffleboard/app/plugin/PluginLoader.java#L36
https://github.com/wpilibsuite/shuffleboard/blob/1197c65e2e6a283122ab73f5524e1da5b8b77708/app/src/main/java/edu/wpi/first/shuffleboard/app/sources/recording/Playback.java#L37
https://github.com/wpilibsuite/shuffleboard/blob/b2841e0686dd858be2bbfdddfa123095134f8f9f/api/src/main/java/edu/wpi/first/shuffleboard/api/sources/recording/Serialization.java#L28
https://github.com/wpilibsuite/shuffleboard/blob/41e70dcb8578f046bd2c726d7933429bcd30fdc1/app/src/main/java/edu/wpi/first/shuffleboard/app/MainWindowController.java#L76

api/src/main/java/edu/wpi/first/shuffleboard/api/widget/ComponentType.java
AustinShalit: What is a Java Type?

app/src/main/java/edu/wpi/first/shuffleboard/app/components/Tile.java
nightpool: Tried doing this but it involved a lot of ugly casts. Probably isn't necessary.

PR #167, opened 2017-08-28
Add sources and widgets for CameraServer streams

SamCarlberg:
Still to do:

- [ ] Recording/playback support
- [ ] Add controls for resolution, brightness, contrast, exposure, saturation
- [x] Multiplatform support (currently hardcoded for Windows 64-bit)
- [x] Properly resize the images with the widget


plugins/cameraserver/src/main/java/edu/wpi/first/shuffleboard/plugin/cameraserver/source/CameraServerSource.java
```diff
@@ -0,0 +1,152 @@
...
+
+  /**
+   * Creates a new JavaFX image that contains the same information as an OpenCV image.
+   *
+   * @param mat the OpenCV image that should be converted
+   */
+  private Image toJavaFxImage(Mat mat) {
+    MatOfByte mob = new MatOfByte();
+    Imgcodecs.imencode(".bmp", mat, mob); // bmp for lossless conversion; the stream is already compressed enough!
+    return new Image(new ByteArrayInputStream(mob.toArray()));
```
JLLeitschuh: Do you want to free the `mob` before exiting this method? This looks like a potential place for a memory leak.

SamCarlberg: I'll make it a field. That'll let OpenCV do the memory management with `imencode`

plugins/cameraserver/src/main/java/edu/wpi/first/shuffleboard/plugin/cameraserver/source/CameraServerSource.java
```diff
@@ -0,0 +1,152 @@
...
+import javafx.beans.binding.BooleanBinding;
+import javafx.scene.image.Image;
+
+public final class CameraServerSource extends AbstractDataSource<CameraServerData> {
+
+  private static final Logger log = Logger.getLogger(CameraServerSource.class.getName());
+
+  private static final Map<String, CameraServerSource> sources = new HashMap<>();
+
+  public static final ITable rootTable = NetworkTable.getTable("/CameraPublisher");
```
JLLeitschuh: Won't this automatically start Network tables? Is this what you want?

plugins/cameraserver/src/main/java/edu/wpi/first/shuffleboard/plugin/cameraserver/source/CameraServerSource.java
JLLeitschuh: How will that help?

SamCarlberg: It won't be a memory leak if there's no memory reallocations

JLLeitschuh: Fair enough. I really dislike the pattern but you have to do ugly things with code when memory management is involved 🤕

plugins/cameraserver/src/main/java/edu/wpi/first/shuffleboard/plugin/cameraserver/source/CameraServerSource.java
SamCarlberg: I think it's fine since this plugin cannot function without network tables

api/src/main/java/edu/wpi/first/shuffleboard/api/components/ResizableImageView.java
```diff
@@ -0,0 +1,67 @@
...
+    return true;
+  }
+
+  @Override
+  public void resize(double width, double height) {
+    setFitWidth(width);
+    setFitHeight(height);
+  }
+
+}
```
AustinShalit: 😱 Why does checkstyle not find this EOL at EOF

plugins/cameraserver/src/main/java/edu/wpi/first/shuffleboard/plugin/cameraserver/CameraServerPlugin.java
```diff
@@ -0,0 +1,69 @@
...
+        "1.0.0",
+        "Provides sources and widgets for viewing CameraServer MJPEG streams.");
+  }
+
+  @Override
+  public void onLoad() {
+    log.info("OpenCV version: " + Core.VERSION);
+    // Make sure the JNI is loaded. If it's not, this plugin can't work!
+    // Calling a function from CameraServerJNI will extract the OpenCV JNI dependencies and load them
+    CameraServerJNI.getHostname();
```
nightpool: I think this is fine to merge as-is, but we should probably file a CameraServer bug about the API gap here.

plugins/cameraserver/src/main/java/edu/wpi/first/shuffleboard/plugin/cameraserver/source/CameraServerSource.java
nightpool: Seems bad to have in a static initializer but we're already doing it for LiveWindow and NetworkTableUtils so ¯\_(ツ)_/¯

plugins/cameraserver/src/main/java/edu/wpi/first/shuffleboard/plugin/cameraserver/data/CameraServerData.java
```diff
@@ -0,0 +1,50 @@
...
+import edu.wpi.first.shuffleboard.api.data.ComplexData;
+
+import java.io.Serializable;
+import java.util.Map;
+
+import javafx.scene.image.Image;
+
+// serialversion UID doesn't matter since this is only used in one JVM at a time
+@SuppressFBWarnings("SE_NO_SERIALVERSIONID")
+public final class CameraServerData extends ComplexData<CameraServerData> implements Serializable {
```
nightpool: This might be general enough that it should be in BasePlugin, so other non-CameraServer sources and widgets could use it

plugins/cameraserver/src/main/java/edu/wpi/first/shuffleboard/plugin/cameraserver/widget/CameraServerWidget.java
```diff
@@ -0,0 +1,36 @@
...
+import org.fxmisc.easybind.EasyBind;
+
+import javafx.fxml.FXML;
+import javafx.scene.image.Image;
+import javafx.scene.image.ImageView;
+import javafx.scene.layout.Pane;
+
+@Description(name = "Camera Stream", dataTypes = CameraServerData.class)
+@ParametrizedController("CameraServerWidget.fxml")
+public class CameraServerWidget extends SimpleAnnotatedWidget<CameraServerData> {
```
nightpool: Like CameraServerData, this might want to be renamed and put into the base plugin (StreamWidget?) so that non-cameraserver streams can also take advantage of it.

(unless the control stuff obviates this?)

plugins/cameraserver/src/main/java/edu/wpi/first/shuffleboard/plugin/cameraserver/data/CameraServerData.java
SamCarlberg: This is why I wanted to bundle the builtin plugins separately, and why I gave them maven-style IDs, so that third parties can depend on our plugins.

plugins/cameraserver/src/main/java/edu/wpi/first/shuffleboard/plugin/cameraserver/data/CameraServerData.java
```diff
@@ -0,0 +1,50 @@
...
+import edu.umd.cs.findbugs.annotations.SuppressFBWarnings;
+import edu.wpi.first.shuffleboard.api.data.ComplexData;
+
+import java.io.Serializable;
+import java.util.Map;
+
+import javafx.scene.image.Image;
+
+// serialversion UID doesn't matter since this is only used in one JVM at a time
+@SuppressFBWarnings("SE_NO_SERIALVERSIONID")
```
AustinShalit: Class is Serializable, but doesn't define serialVersionUID

>This class implements the Serializable interface, but does not define a serialVersionUID field.  A change as simple as adding a reference to a .class object will add synthetic fields to the class, which will unfortunately change the implicit serialVersionUID (e.g., adding a reference to String.class will generate a static field class$java$lang$String). Also, different source code to bytecode compilers may use different naming conventions for synthetic variables generated for references to class objects or inner classes. To ensure interoperability of Serializable across versions, consider adding an explicit serialVersionUID.

This seems like something reasonable to do.

plugins/cameraserver/src/main/java/edu/wpi/first/shuffleboard/plugin/cameraserver/source/CameraServerSource.java
```diff
@@ -0,0 +1,163 @@
...
+  /**
+   * Continuously reads frames from the MJPEG stream as long as the calling thread has not been interrupted. Because
+   * this is a (nearly) infinite loop, this <strong>cannot</strong> be called from the JavaFX application thread. We
+   * call this in an executor service to run asynchronously.
+   */
+  private void grabForever() {
+    if (Platform.isFxApplicationThread()) {
+      throw new IllegalStateException("This may not run on the FX application thread!");
+    }
+    final Thread thread = Thread.currentThread();
```
AustinShalit: Imo this is clearer if we do not cache the thread.  If we are going to cache this, we should rename the variable to be currentThread.

plugins/cameraserver/src/main/java/edu/wpi/first/shuffleboard/plugin/cameraserver/data/CameraServerData.java
JLLeitschuh: @SamCarlberg Are you arguing that because the object will never need to be serialized and read in by another JVM (even from another launch) this isn't necessary?

SamCarlberg:
I'd appreciate if someone with a roborio could test this with recording and playback

api/src/main/java/edu/wpi/first/shuffleboard/api/sources/recording/Recorder.java
```diff
@@ -124,4 +132,7 @@ public void setRunning(boolean running) {
     this.running.set(running);
   }
 
+  public File getRecordingFile() {
```
JLLeitschuh: `Optional<File>` as it can be null?

nightpool:
Widget is fine, recording is flaky (can't payback in shuffleboard, video files often corrupt).

Good to merge after marking cameras as non-recordable

PR #120, opened 2017-08-09
Team Connect

AustinShalit:
2 Problems I need to solve:
- [ ] Dialog starts small (#124)
- [x] when changing the port number, NetworkTables does not start again (I am calling initialize but it does not start).  Update: This is a bug in ntcore.

PR #168, opened 2017-08-28
Fix prefs dialog not having stylesheets set

SamCarlberg:
Caused by #120


AustinShalit:
Oops.  I know that you commented on the PR to fix it and I did fix it.  Must have been removed by a rebase.

PR #174, opened 2017-09-07
Make text field widgets editable

### LINKED ISSUE #47: TextView widgets are not editable
See title

SamCarlberg:
Fixes #47

ATM allows for boolean text entries (`"true"` is the `true` value, anything else becomes `false`). I think that's acceptable since checkboxes/toggle buttons/toggle switches all exist and are incredibly easier to use

### DIFF
--- a/plugins/base/src/main/java/edu/wpi/first/shuffleboard/plugin/base/widget/TextView.java
+++ b/plugins/base/src/main/java/edu/wpi/first/shuffleboard/plugin/base/widget/TextView.java
@@ -1,16 +1,17 @@
 package edu.wpi.first.shuffleboard.plugin.base.widget;
 
-import edu.wpi.first.shuffleboard.api.sources.DataSource;
+import edu.wpi.first.shuffleboard.api.components.NumberField;
 import edu.wpi.first.shuffleboard.api.widget.Description;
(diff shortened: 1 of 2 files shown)


plugins/base/src/main/java/edu/wpi/first/shuffleboard/plugin/base/widget/TextView.java
```diff
@@ -26,23 +27,36 @@
...
   private Pane root;
+  @FXML
+  private TextField textField;
+  @FXML
+  private NumberField numberField;
+
+  @FXML
+  private void initialize() {
+    text.bind(EasyBind.map(dataProperty(), this::simpleToString));
+    MonadicBinding<Boolean> isNumber = EasyBind.map(dataProperty(), d -> d instanceof Number).orElse(false);
```
nightpool: does this need to be retained for GC reasons?

SamCarlberg: No. This is on the stack so it's gone. I was originally planning in using those on both text fields, but `MonadicBinding<Boolean>`doesn't support .not() like a BooleanBinding

Issue #43, opened 2017-07-19
Recording file should not be .frc


AustinShalit:
This tool can be used for things other than frc.

SamCarlberg:
Agreed. @nightpool thoughts?

nightpool:
Yes this was a review comment I had on the original PR that we didn't fix.

.sblog maybe?

nightpool:
`.sbr` 

@rrlapointe to take and work on

AustinShalit:
@rrlapointe `.sbr`?

SamCarlberg:
Closed by #179

PR #179, opened 2017-09-07
Change recording file extension from .frc to .sbr

rrlapointe:
Change the file extension used for recordings from `.frc` to `.sbr`


rrlapointe:
This PR resolves #43

PR #180, opened 2017-09-08
Improve Findbugs Error Reporting by using emacs output instead of XML

JLLeitschuh:
This prints out much saner output that is far easier to read.

No more trying to read incomprehensible XML!


nightpool:
looks like spotless is complaining.

JLLeitschuh:
There, that's the same code that I'm using at Plexxi and it works for me.

JLLeitschuh:
Thoughts? Can this be merged?

PR #184, opened 2017-09-17
Add Number Bar widget

SamCarlberg:
Also allows SimpleAnnotatedWidgets to use the data type's default value when no source is present, to eliminate annoying NPEs when widgets are initialized


nightpool:
wasn't the consensus to also remove Number Slider? why keep it around if it's just going to confuse people? 

Also the name "Linear Indicator" is very unintuitive.

AustinShalit:
>Also the name "Linear Indicator" is very unintuitive.

Graduated bar?

nightpool:
I honestly would just rename this one to "Number Bar" personally. I think the consensus is to remove the "make default for numbers" part of this pull request though, right?

PR #185, opened 2017-09-17
Fix #124

### LINKED ISSUE #124: Tab Preferences dialog starts too small
It starts with scroll bars now.  It should not.

nightpool:
When PropertySheet has padding, it doesn't calculate it's height correctly.
Dialogs give some padding to their content
So PropertySheets directly embedded in Dialogs are too short


SamCarlberg:
This should also be used in the app preferences dialog. And can you upload screenshots?

nightpool:
I'm not sure what the usefulness of screenshots are for a one-line change?

Will fix AppPreferences.

nightpool:
image

image

PR #191, opened 2017-09-19
Change the default number widget type from NumberSlider to TextView

### LINKED ISSUE #188: Making the default number field be a "number field"
Currently the default number field is a slider, although I'd expect that most people would want to see the actual number. We probably should change this behavior.

rrlapointe:
Resolves #188 
Not totally sure if TextView is acceptable or if a new widget is needed; the javadoc for TextView claims it is for text, numbers, and booleans.


AustinShalit:
I think the correct UI element for numbers should be a [spinner](https://docs.oracle.com/javase/8/javafx/api/javafx/scene/control/Spinner.html) but we can implement that later.

What happens when you try to enter nonnumerical characters (A-Z) into the field right now?

rrlapointe:
From my experimentation, the TextView widget seems to be read-only. I cannot enter anything into it; it only displays data from NetworkTables.

SamCarlberg:
@rrlapointe see #174

rrlapointe:
Thanks @SamCarlberg I will see if there are any problems when combining that PR and this one

nightpool:
@AustinShalit #174 adds code for that case so I'm fine merging this

PR #194, opened 2017-09-19
Remove progress bar widget

### LINKED ISSUE #84: remove progress bar widget
negative vs positive values should be clear from a glance

image

rrlapointe:
In conjunction with #184, resolves #84

Can someone confirm that removing the CSS styles will not affect anything else? I looked around the UI and don't believe that they apply to anything besides the progress bar widget.


SamCarlberg:
The CSS shouldn't be changed because progress bars may be used elsewhere and we want to keep a consistent LnF

rrlapointe:
Ok, reverted removing the CSS

PR #196, opened 2017-09-20
Bump to stable release of Junit 5

JLLeitschuh:
Also remove a gradle API workarround that was actually my own mistake.

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -11,7 +11,7 @@ buildscript {
         mavenCentral()
     }
     dependencies {
-        classpath("org.junit.platform:junit-platform-gradle-plugin:1.0.0-RC2")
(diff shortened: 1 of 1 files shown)


build.gradle.kts
```diff
@@ -196,16 +196,6 @@ task<Wrapper>("wrapper") {
 }
 
 /**
- * Workaround fix for calling [org.gradle.testing.jacoco.plugins.JacocoPluginExtension.applyTo]
- *
- * [Issue details here](https://github.com/gradle/kotlin-dsl/issues/458)
- */
-fun org.gradle.testing.jacoco.plugins.JacocoPluginExtension.applyToHelper(task : Task) {
```
JLLeitschuh: In hindsight this was simply because I was passing the wrong type to this method. There was an actual bug in the Gradle API that has been fixed, but I conflated the two issues.

PR #198, opened 2017-09-21
Fix tile titles

### LINKED ISSUE #29: Widget titlebar probably should default to the root name
The titlebar on widgets currently has the full path for the value. So a typical value would be "SmartDashboard/name". With small widgets, the name isn't shown because "SmartDashboard" fills the entire title area. Maybe show the full value when mousing over or in a properties dialog for the widget.

SamCarlberg:
Sets the text overrun to the left side to make titles look like `...system/Victor/Speed` instead of `/LiveWindow/MySub...`

Uses the tab source prefix to remove redundant leading information

Changing the widget title no longer changes the name of its source

Fixes #29

### DIFF
--- a/api/src/main/java/edu/wpi/first/shuffleboard/api/components/EditableLabel.java
+++ b/api/src/main/java/edu/wpi/first/shuffleboard/api/components/EditableLabel.java
@@ -48,7 +48,6 @@ public class EditableLabel extends StackPane {
       if (isEditing) {
         editField.requestFocus();
       }
-      label.setVisible(!isEditing);
     });
   }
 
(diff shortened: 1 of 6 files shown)


api/src/main/java/edu/wpi/first/shuffleboard/api/components/EditableLabel.java
```diff
@@ -48,7 +48,6 @@ public EditableLabel() {
       if (isEditing) {
         editField.requestFocus();
       }
-      label.setVisible(!isEditing);
```
nightpool: ?

SamCarlberg: The visible property is already bound so all this does is throw an exception

app/src/main/java/edu/wpi/first/shuffleboard/app/WidgetPaneController.java
```diff
@@ -252,6 +260,23 @@ private void setupTile(Tile tile) {
...
+        // Prefix without the protocol or leading slash
+        String namePrefix = SourceTypes.getDefault().stripProtocol(tab.getSourcePrefix()).replaceFirst("^/", "");
+        tile.getContent().allWidgets()
+            .filter(w -> w.getTitle().startsWith(namePrefix))
+            .forEach(w -> w.setTitle(w.getTitle().substring(namePrefix.length())));
+      } catch (NoSuchFieldException | IllegalAccessException e) {
+        log.log(Level.SEVERE, "Could not get the tab containing the widget pane!", e);
+      }
+    }
+
```
nightpool: from discussion: let's leave this off for now until we can refactor auto-population

PR #206, opened 2017-09-28
Add graph widget

SamCarlberg:
Broken on playback scrubbing or when playback loops

### DIFF
--- a/app/src/main/java/edu/wpi/first/shuffleboard/app/WidgetTileController.java
+++ b/app/src/main/java/edu/wpi/first/shuffleboard/app/WidgetTileController.java
@@ -1,14 +1,11 @@
 package edu.wpi.first.shuffleboard.app;
 
-import edu.wpi.first.shuffleboard.api.sources.DataSource;
 import edu.wpi.first.shuffleboard.api.widget.Widget;
 import edu.wpi.first.shuffleboard.app.components.WidgetPropertySheet;
 import edu.wpi.first.shuffleboard.app.components.WidgetTile;
 import edu.wpi.first.shuffleboard.app.sources.DestroyedSource;
 
-import org.fxmisc.easybind.EasyBind;
-import org.fxmisc.easybind.monadic.PropertyBinding;
-
+import javafx.beans.InvalidationListener;
 import javafx.beans.binding.Bindings;
 import javafx.css.PseudoClass;
 import javafx.fxml.FXML;
@@ -23,9 +20,6 @@ public class WidgetTileController {
   @FXML
   private Label titleLabel;
 
-  // Store as a field to prevent GC
-  private PropertyBinding<DataSource> sourceProperty; //NOPMD could be a local variable
-
   /**
    * Pseudoclass used on tiles when the widget inside loses its source.
    */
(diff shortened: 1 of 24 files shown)


app/src/main/java/edu/wpi/first/shuffleboard/app/WidgetTileController.java
```diff
@@ -34,20 +28,31 @@
...
-        tile.pseudoClassStateChanged(NO_SOURCE, true);
-        tile.setDisable(true);
-      } else {
-        tile.pseudoClassStateChanged(NO_SOURCE, false);
-        tile.setDisable(false);
-      }
-    });
+  }
+
+  private void setTitleFor(Widget widget) {
```
nightpool: this setTileFor stuff needs to be re-worked for Widgets that control their own titles using a titleProperty

app/src/main/java/edu/wpi/first/shuffleboard/app/components/DashboardTabPane.java
```diff
@@ -205,8 +205,8 @@ private void populate() {
           if (source.getDataType() != DataTypes.Unknown
               && source.getDataType() != DataTypes.Map
               && !Widgets.getDefault().widgetNamesForSource(source).isEmpty()) {
-            Widgets.getDefault().createWidget(Widgets.getDefault().widgetNamesForSource(source).get(0), source)
-                .ifPresent(w -> getWidgetPane().addWidget(w));
+            Widgets.getDefault().pickWidgetNameFor(source.getDataType())
+                .ifPresent(w -> getWidgetPane().addWidget(Widgets.getDefault().createWidget(w, source).get()));
```
nightpool: why not `map(w -> createWidget(w, source)).ifPresent(w -> addWidget(w))`?

app/src/main/java/edu/wpi/first/shuffleboard/app/components/DashboardTabPane.java
```diff
@@ -224,8 +224,8 @@ private boolean shouldAutopopulate(String sourceId) {
      */
     private boolean noExistingWidgetsForSource(String id) {
       for (WidgetTile tile : getWidgetPane().getTiles()) {
-        DataSource<?> source = tile.getWidget().getSource();
-        if (id.startsWith(source.getId())) {
+        boolean match = tile.getWidget().getSources().stream().anyMatch(s -> id.startsWith(s.getId()));
```
nightpool: this whole function can be

`getTiles.stream().allMatch(tile -> !tile.getWidget().getSource.stream().anyMatch(s -> id.startsWith(s.getId()))`

checkstyle.xml
```diff
@@ -109,7 +109,7 @@
                      value="Type name ''{0}'' must match pattern ''{1}''."/>
         </module>
         <module name="MemberName">
-            <property name="format" value="^[a-z][a-z0-9][a-zA-Z0-9]*$"/>
+            <property name="format" value="^[a-z][a-zA-Z0-9]*$"/>
```
nightpool: Why is this changed? This is part of the google guide afaik and forbids member names like `aThing`

nightpool: ah, xAxis and yAxis. Should probably just be changed to have the the `x|y` (`|z`?) exceptions from LocalVariableName

AustinShalit:
Please rebase with the latest master.  frcjenkins **must** pass now.

SamCarlberg:
Updated to master.

Dragging sources onto tiles with graphs also seems to behave erratically. I'll have to look into that before merging; it's a pain in the butt to add extra sources to graph widgets as-is

Caveats:

1. During playback, graphs only behave properly when playback is automatically running. Scrubbing or pausing throws it way off
2. Updating graphs use a lot of CPU power.

2 has been fixed in Java 10 as far as I can tell, but that won't be released until March. We may have to copy the entire graph class hierarchy into shuffleboard to get the performance in line; I believe graphs make heavy use of package-private APIs

PR #220, opened 2017-10-03
Add a test that launches the whole application

JLLeitschuh:
Adds a new project so that this test will be launched in a separate process.

This is the contents of the README.Travis.md that I've included in this PR

------

# Travis CI

By default Travis uses Ubuntu 14.04 LTS for building,
but it is possible to run the build in any Linux distribution using [Docker](https://www.docker.com/).

Network Tables relies upon the ABI (Application Binary Interface) that is incompatible
with Ubuntu 14.04. As such, trying to call any Network Tables methods cause errors
when the JNI library is loaded.

The workaround for this issue is to use a Docker container running Ubuntu 16.04
inside of Travis CI.

This build configuration is inspired by the
[openSUSE/snapper](https://github.com/openSUSE/snapper) project.

## Setup
The `Dockerfile` defines the steps needed for building the Docker image that
Shuffleboard is tested within.

The `.travis.yml` file defines the build matrix which runs the build using Docker.

## Running the Build Locally

**NOTE** This is only necessary if you are having issues with tests failing on TravisCI.
You should be able to develop for Shuffleboard on your local machine without involving
Docker.

 1. [Install and start Docker](https://docs.docker.com/engine/installation/)
 2. Run a similar same commands as in the `.travis.yml` file:
    1. First build the docker image locally:
       ```bash
       docker build -f Dockerfile -t shuffleboard-devel .
       ```
       The Docker image automatically includes also the copy of the current Shuffleboard sources.
       If you change any of the sources you will need to re-run this step.
       Don't worry, it gets faster after the first time as Docker can skip downloading the
       `apt` packages.
    2. Then run the build:
       ```bash
       docker run -it --rm shuffleboard-devel ./.travis.ubuntu.sh
       # ALTERNATIVELY:
       # If you want to use the gradle dependency cache on your host machine:
       docker run -it --rm -v $HOME/.gradle/:/root/.gradle/ shuffleboard-devel ./.travis.ubuntu.sh
       ```
       (The `--rm` will cleanup the new Docker image layer created by the build,
       if you want to inspect the build artifacts then remove it.)
 3. If you need to debug a failure then run this instead of the Travis script and run the
    build steps manually:
    ```bash
    docker run -it --rm shuffleboard-devel bash
    ```
    If you need an editor or some other tool you can install them via the respective packaging tool,
    see the Dockerfile for examples.


.travis.yml
```diff
@@ -1,4 +1,8 @@
-# NOTE: Don't use `sudo` anywhere or the build will run on Travis's slower, outdated build system
```
JLLeitschuh: Can't get arround this one because we want to use docker. Sad to see it go.

Dockerfile
```diff
@@ -0,0 +1,18 @@
...
+  apt-get install -y oracle-java8-installer && \
+  rm -rf /var/lib/apt/lists/* && \
+  rm -rf /var/cache/oracle-jdk8-installer
+
+ENV JAVA_HOME /usr/lib/jvm/java-8-oracle
+
+RUN mkdir -p /usr/src/app
+WORKDIR /usr/src/app
+
+COPY . /usr/src/app
```
PeterJohnson: EOL at EOF

.travis.ubuntu.sh
```diff
@@ -0,0 +1,10 @@
+#! /bin/bash
+
+set -e -x
+
+Xvfb :99 & export DISPLAY=:99
+
+./gradlew --version --console=plain
+./gradlew spotlessCheck check jacocoJunit5TestReport --stacktrace --console=plain -PlogTests -Pheadless
+
+bash <(curl -s https://codecov.io/bash)
```
PeterJohnson: EOL at EOF

JLLeitschuh: You know what, you are absolutely right. I'm going to add a spotless rule to gradle about this.

JLLeitschuh: This should be forever resolved in this file because of this new spotless check:
https://github.com/wpilibsuite/shuffleboard/pull/220/files#diff-392475fdf2bc320d17762ed97109a121R42

.gitattributes
```diff
@@ -1 +1,3 @@
 *.kts text eol=lf
+*.yml text eol=lf
+Dockerfile text eol=lf
```
JLLeitschuh: Necessary because of windows.

JLLeitschuh:
@nightpool @SamCarlberg @AustinShalit This PR is "complete" in my book. Feel free to merge if you want or leave comments and I'll clean it up more.

Dockerfile
JLLeitschuh: Fixed! Thanks!

PR #225, opened 2017-10-05
Remove print when saving layout file

SamCarlberg:


SamCarlberg:
Merging with failing appveyor due to spurious test failure. Travis and Jenkins both pass.

Issue #175, opened 2017-09-07
Toggle Switches in Tile property sheets are invisible


SamCarlberg:
This is a problem with this section in `base.css`

```css
.tile .property-sheet * {
    /* TODO figure out exactly what needs to be set */
    -fx-background-color: transparent;
}
```

This section is intended to make the background of the sheet transparent to match the background color of the tile.

If we're going to show tile preferences in dialog windows, this is a non-issue

PR #228, opened 2017-10-11
Added property editor dialog as context menu item for Widgets

### LINKED ISSUE #212: Property sheet for widgets should be on the context menu
When double-clicking on a widget (a somewhat non intuitive operation) the widget area is replaced by it's property sheet. If the widget is small, the property sheet is unusable requiring the widget to be resized, then the property sheet edited, then double-click again, then the, now larger widget needs to be reduced to it's original size.

Instead, the property sheets should just be dialog boxes that pop up on a "Properties..." context menu entry. Then no resizing is required, and users don't have to learn the double-click gesture.

Octogonapus:
Closes #212

image

image

### DIFF
--- a/app/src/main/java/edu/wpi/first/shuffleboard/app/components/WidgetTile.java
+++ b/app/src/main/java/edu/wpi/first/shuffleboard/app/components/WidgetTile.java
@@ -4,16 +4,13 @@ import edu.wpi.first.shuffleboard.api.sources.DataSource;
 import edu.wpi.first.shuffleboard.api.widget.TileSize;
 import edu.wpi.first.shuffleboard.api.widget.Widget;
 import edu.wpi.first.shuffleboard.app.sources.DestroyedSource;
-
-import org.fxmisc.easybind.EasyBind;
-import org.fxmisc.easybind.monadic.PropertyBinding;
-
 import javafx.beans.binding.Bindings;
 import javafx.beans.property.BooleanProperty;
 import javafx.beans.property.SimpleBooleanProperty;
 import javafx.css.PseudoClass;
 import javafx.scene.Node;
-import javafx.scene.input.MouseEvent;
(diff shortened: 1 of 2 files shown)


app/src/main/java/edu/wpi/first/shuffleboard/app/components/WidgetTile.java
```diff
@@ -54,35 +51,18 @@ private WidgetTile() {
...
-    if (newV != null) {
-      newV.addEventHandler(MouseEvent.MOUSE_CLICKED, event -> {
-        if (event.getClickCount() == 2) {
-          toggleShowWidget();
-        }
-      });
-    }
-  }
-
   private Node createCenter() {
```
SamCarlberg: This method can be removed now that the tile can only contain the widget

Octogonapus:
This probably closes #175

app/src/main/java/edu/wpi/first/shuffleboard/app/components/WidgetTile.java
```diff
@@ -100,5 +80,4 @@ public void setShowWidget(boolean showWidget) {
   public void toggleShowWidget() {
```
SamCarlberg: This method (and everything else related to the `showWidget` property) can be removed now that the tile can only contain the widget

SamCarlberg:
This doesn't close #175 since the property sheets are used elsewhere

bradamiller:
It pops up an empty dialog box if there are no properties, for example, with a text field. Also, there is no title on the dialogBox. Should at least have something static like "Edit widget properties". Otherwise seems to work.

It changes the widget properties as you type. So if you change a sliders limits, you see it changing as you are typing. I wonder if sometime (not now, there are more important things to do first), it should have ok, cancel, and apply buttons. Then it might save the properties at the start and put them back on cancel.

PR #232, opened 2017-10-11
Add a connection indicator 

### LINKED ISSUE #54: Add NetworkTable connection indicator
I think this should be built into the UI (like in the title bar) and not a widget.

AustinShalit:
closes #54


AustinShalit:
Welp... started at the wrong commit.  Going to rebase

SamCarlberg:
I can't get this to work when connecting/disconnecting from a local server (outlineviewer). The label always says "Connected to 127.0.0.1" even when I change the remote ID to something like "190" or "8.8.8.8". This might be a problem with network tables not properly updating the server

I've already fixed this on a different branch locally. However, I'm worried about binding too closely to network tables in the app. This should belong in the NetworkTables plugin

app/src/main/resources/edu/wpi/first/shuffleboard/app/PlaybackControls.fxml
```diff
@@ -8,44 +8,55 @@
...
+                    <ImageView>
+                        <Image url="/edu/wpi/first/shuffleboard/app/icons/icons8-Fast Forward-16.png"/>
+                    </ImageView>
+                </graphic>
+            </Button>
+            <HBox alignment="CENTER" spacing="4">
+                <Scrubber fx:id="progressScrubber"/>
+                <ToggleSwitch text="Loop" fx:id="loopingSwitch" alignment="CENTER" textAlignment="CENTER"/>
+                <Label fx:id="progressLabel"/>
+            </HBox>
```
nightpool: This needs to be in a separate component, putting it as part of the PlaybackIndicator doesn't make any sense architecturally.

SamCarlberg:
@AustinShalit I think this could be better implemented as a data type + widget in the NetworkTables plugin. It'd behave similarly to the old dashboard in that regard, as well as not tie the app too strongly to NetworkTables

sciencewhiz:
SmartDashboard originally only had a widget, but now also shows connection status in the titlebar. 

image

SamCarlberg:
Superseded by #324

PR #233, opened 2017-10-11
Do not autopopulate metadata

SamCarlberg:
No widgets will be added for metadata sources.


bradamiller:
It's much better!

Is it possible to just have the names and values shown as one item for each pair. So it would have a label like "AnalogGyro 1" and value would be above it. Rather than showing "AnalogGyro 1/Name, /Subsystem, /value. Just 53.9 over "AnalogGyro 1"

SamCarlberg:
That's going to be a separate PR

PR #237, opened 2017-10-12
Add widget for PID Controllers

SamCarlberg:


SamCarlberg:
@AustinShalit thoughts on loosening up checkstyle? I don't like having `@SuppressWarnings` everywhere I use PIDF constants

plugins/base/src/main/java/edu/wpi/first/shuffleboard/plugin/base/data/PIDControllerData.java
```diff
@@ -0,0 +1,148 @@
+package edu.wpi.first.shuffleboard.plugin.base.data;
+
+import com.google.common.collect.ImmutableMap;
+
+import edu.wpi.first.shuffleboard.api.data.ComplexData;
+
+import java.util.Map;
+import java.util.Objects;
+
+public final class PIDControllerData extends ComplexData<PIDControllerData> {
```
SamCarlberg: Classes like like this are _really_ tempting to use Lombok for:

```java
@Value
@Wither
@AllArgsConstructor
public class PIDControllerData extends ComplexData<PIDControllerData> {

  private final double p;
  private final double i;
  private final double d;
  private final double f;
  private final double setpoint;
  private final boolean enabled;

  public PIDControllerData(Map<String, Object> map) {
    this(...);
  }

  @Override
  public Map<String, Object> asMap() {
    return ...
  }

}
```

Removes ~90 lines of boilerplate constructors/getters/withers/equals/hashcode/toString

JLLeitschuh: Or a kotlin data class.

PR #243, opened 2017-10-14
Make app preferences persistent

### LINKED ISSUE #169: App preferences should be saved
They're not currently saved and need to be re-entered every time the app is launched.

This could be implemented on a save-file level to ensure that saved dashboards will be identical across different computers, letting them be easily shared between team members, or on a per-user basis so that each user will have the same settings across all projects.

SamCarlberg:
Closes #169


SamCarlberg:
Superseded by #248

PR #246, opened 2017-10-15
Make buttons and toggle buttons look more like raised buttons

### LINKED ISSUE #186: Buttons don't follow Material Design guidelines
For example, these buttons aren't obviously buttons:

image

Buttons should have theme-appropriate colorless backgrounds, drop shadow, and colored text. Hover and pressed should have slightly darker backgrounds.

See: https://material.io/guidelines/components/buttons.html

SamCarlberg:
Also fixes ToggleButton widgets not showing the source name in the button, along with adding some extra padding to make them easier to tell apart from parent tiles

Closes #186 

# Screenshots

## Light theme
buttons-light

## Dark theme
buttons-dark

### DIFF
--- a/api/src/main/resources/edu/wpi/first/shuffleboard/api/material.css
+++ b/api/src/main/resources/edu/wpi/first/shuffleboard/api/material.css
@@ -119,11 +119,17 @@
 
 .toggle-button, .button {
     -fx-text-fill: -fx-text-button-normal;
-    -fx-font-family: 'Roboto';
     -fx-font-weight: bold;
     -fx-background-insets: 0.0;
-    -fx-background-radius: 4.0;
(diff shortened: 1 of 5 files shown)


nightpool:
is it just me, or is the shadow *lighter* when it's clicked? I'm worried it won't look interactive enough in the "on" state.

SamCarlberg:
That's probably a contrast issue. The shadow's color doesn't change.

nightpool:
Okay.

this PR breaks the inline buttons in motor controllers:

image

---------

(before)

image

SamCarlberg:
Ah, that widget never had spacing set in the controls. Wasn't a problem before because the button was basically transparent. Just fixed that.

PR #248, opened 2017-10-17
Add plugin properties, make them persistent

### LINKED ISSUE #169: App preferences should be saved

SamCarlberg:
Supersedes #243 
Closes #169 

Moves network table server property to the NetworkTables plugin  
Fixes a bug with where changing the NetworkTable server property wouldn't connect to the new server

Update dialog CSS to make the content span the width of the dialog. This gives a nice solid bar for the preference tabs:

shuffleboard-1


api/src/main/java/edu/wpi/first/shuffleboard/api/plugin/Plugin.java
```diff
@@ -134,6 +135,16 @@ public void onUnload() {
   }
 
   /**
+   * Gets a list of properties of this plugin that can be changed by users. Properties that are sensitive to rapid
+   * changes (for example, a server URI that will attempt a connection on a change) should be wrapped in a
+   * {@link edu.wpi.first.shuffleboard.api.prefs.FlushableProperty FlushableProperty} to ensure that a change will only
+   * occur when a user manually confirms the change.
```
JLLeitschuh: Include a code sample about how to do that??

plugins/networktables/src/main/java/edu/wpi/first/shuffleboard/plugin/networktables/NetworkTablesPlugin.java
```diff
@@ -71,4 +120,12 @@ public void onUnload() {
     );
   }
 
+  @Override
+  public List<Property<?>> getProperties() {
+    return ImmutableList.of(
+        // use FlushableProperty so changes made are only effected when committed by the user
+        new FlushableProperty<>(serverId)
```
JLLeitschuh: Expose a static creation method for `FlushableProperty` instead of a constructor??

SamCarlberg: Something like `PropertyUtils.makeFlushable(...)`?

But why?

JLLeitschuh: No, I'm suggesting:

`FlushableProperty.from(...)`

Reasoning:

> # Item 1: Consider static factory methods instead of constructors
>
> A class can provide a public static factory method, which is simply a static method that returns an instance of the class.
> ```
> public static Boolean valueOf(boolean b) {
>     return b ? Boolean.TRUE : Boolean.FALSE;
> }
> ```
> Providing a static factory method instead of a public constructor has both advantages and disadvantages.
>
> Advantages:
> 1.    Unlike constructors, they have names.
> 2.    Unlike constructors, they are not required to create a new object each time they’re invoked.
> 3.    Unlike constructors, they can return an object of any subtype of their return type.
> 4.   They reduce the verbosity of creating parameterized type instances.
>
> ```
> Map<String, List<String>> m = new HashMap<String, List<String>>();
> ```
> 
> With static factories, however, the compiler can figure out the type parameters for you. This is known as type inference. For example, suppose that HashMap provided this static factory:
>
> ```
> public static <K, V> HashMap<K, V> newInstance() {
>     return new HashMap<K, V>();
> }
> ```
>
> Then you could replace the wordy declaration above with this succinct alternative:
>
> `Map<String, List<String>> m = HashMap.newInstance();`
>
> Disadvantages:
> 1. Classes without public or protected constructors cannot be subclassed.
> 2. They are not readily distinguishable from other static methods.
>
> They do not stand out in API documentation in the way that constructors do, so it can be difficult to figure out how to instantiate a class that provides static factory methods instead of constructors.
>
> Here are some common names for static factory methods:
> • valueOf—Returns an instance that has, loosely speaking, the same value as its
parameters. Such static factories are effectively type-conversion methods.
> • of—A concise alternative to valueOf, popularized by EnumSet (Item 32).
>• getInstance—Returns an instance that is described by the parameters but
cannot be said to have the same value. In the case of a singleton, getInstance
takes no parameters and returns the sole instance.
> • newInstance—Like getInstance, except that newInstance guarantees that
each instance returned is distinct from all others.
> • getType—Like getInstance, but used when the factory method is in a different
class. Type indicates the type of object returned by the factory method.
> • newType—Like newInstance, but used when the factory method is in a different
class. Type indicates the type of object returned by the factory method.
>
> In summary, static factory methods and public constructors both have their
uses, and it pays to understand their relative merits. Often static factories are preferable,
so avoid the reflex to provide public constructors without first considering
static factories.
> \-[Effective Java 2nd Edition by Joshua Bloch](http://jtechies.blogspot.com/2012/07/item-1-consider-static-factory-methods.html)

SamCarlberg: I see no need for it

JLLeitschuh: Its part of your publicly exposed API. I would highly recommend it.

PR #250, opened 2017-10-21
Add relay widget

rrlapointe:
Add a relay widget with buttons for On, Off, Forward, and Reverse.

screenshot_2017-10-21_18-28-17


plugins/base/src/main/java/edu/wpi/first/shuffleboard/plugin/base/widget/RelayWidget.java
```diff
@@ -0,0 +1,101 @@
...
+
+import edu.wpi.first.shuffleboard.api.widget.Description;
+import edu.wpi.first.shuffleboard.api.widget.ParametrizedController;
+import edu.wpi.first.shuffleboard.api.widget.SimpleAnnotatedWidget;
+import javafx.fxml.FXML;
+import javafx.scene.layout.Pane;
+
+@Description(name = "Relay Widget", dataTypes = String.class)
+@ParametrizedController("RelayWidget.fxml")
+public class RelayWidget extends SimpleAnnotatedWidget<String> {
```
SamCarlberg: We should have a data type for relays (see the `...plugin.base.data` package). That allows us to use this widget to autopopulate relays

plugins/base/src/main/java/edu/wpi/first/shuffleboard/plugin/base/widget/RelayWidget.java
```diff
@@ -0,0 +1,101 @@
...
+          onButton.setSelected(true);
+          break;
+        case FORWARD:
+          forwardButton.setSelected(true);
+          break;
+        case REVERSE:
+          reverseButton.setSelected(true);
+          break;
+        default:
+          throw new AssertionError();
```
SamCarlberg: Need a message

plugins/base/src/main/resources/edu/wpi/first/shuffleboard/plugin/base/widget/RelayWidget.css
```diff
@@ -0,0 +1,7 @@
+.toggle-button {
```
SamCarlberg: File should use standard CSS naming scheme (`relay-widget.css`)

bradamiller:
Ryan... have you had a chance to make the changes indicated above? Are you OK with them?

rrlapointe:
I am planning on making the changes today

plugins/base/src/main/resources/edu/wpi/first/shuffleboard/plugin/base/widget/RelayWidget.fxml
```diff
@@ -0,0 +1,27 @@
+<?xml version="1.0" encoding="UTF-8"?>
+
+<?import javafx.scene.control.ToggleButton?>
+<?import javafx.scene.control.ToggleGroup?>
+<?import javafx.scene.layout.StackPane?>
+<?import javafx.scene.layout.VBox?>
+
+<StackPane fx:id="root" stylesheets="@relay-widget.css" xmlns="http://javafx.com/javafx/8.0.121"
+           xmlns:fx="http://javafx.com/fxml/1"
+           fx:controller="edu.wpi.first.shuffleboard.plugin.base.widget.RelayWidget">
```
SamCarlberg: No minimum size set, makes the widget look odd at small sizes. 128x64 seems appropriate

relay-no-min-size

plugins/base/src/main/resources/edu/wpi/first/shuffleboard/plugin/base/widget/RelayWidget.fxml
```diff
@@ -0,0 +1,27 @@
...
+                          mnemonicParsing="false" text="Off">
+                <toggleGroup>
+                    <ToggleGroup fx:id="state"/>
+                </toggleGroup>
+            </ToggleButton>
+            <ToggleButton id="onButton" fx:id="onButton" maxHeight="Infinity" maxWidth="Infinity"
+                          mnemonicParsing="false" text="On" toggleGroup="$state"/>
+            <ToggleButton id="forwardButton" fx:id="forwardButton" maxHeight="Infinity" maxWidth="Infinity"
+                          mnemonicParsing="false" text="Forward" toggleGroup="$state"/>
+            <ToggleButton id="reverseButton" fx:id="reverseButton" maxHeight="Infinity" maxWidth="Infinity"
```
SamCarlberg: These buttons don't grow vertically, which results in a lot of empty space at large heights

relay-buttons-don t-grow

plugins/base/src/main/resources/edu/wpi/first/shuffleboard/plugin/base/widget/RelayWidget.fxml
SamCarlberg: No minimum size specified. Causes issues at low tile sizes. 128x64 seems like it could be OK

relay-no-min-size

plugins/base/src/main/resources/edu/wpi/first/shuffleboard/plugin/base/widget/RelayWidget.fxml
SamCarlberg: It'd also be nice to have some spacing between buttons... `2` or maybe `4` would work

bradamiller: @sam if all that's left to fix is the buttons growing vertically, which is not clearly the right thing to do, then is it OK to merge now?

rrlapointe:
@SamCarlberg re: 'Select "Off" instead of indeterminate state when untoggling a button':
There was never any indeterminate state because the buttons could not be untoggled. You have made them untoggleable, which I believe is a confusing functionality--if the user wants to turn the relay off, they should press the off button.  I believe the initial behavior--where pressing the already-selected button did nothing--was better.

SamCarlberg:
The problem was that deselecting a button didn't actually change anything. If the "Reverse" button were toggled on, then the relay would go into that state. If that button was then deselected, the relay would stay in the "Reverse" state which the widget wouldn't display.

The change I made was so that deselecting a button would automatically select the "Off" button and turn off the relay.

rrlapointe:
What I am trying to say is that previously, you could not deselect the button. Pressing the button that was currently selected would not deselect it.

PR #255, opened 2017-10-23
Fix several issues with recording and playback

### LINKED ISSUE #152: Source tree previews don't update from playback

SamCarlberg:
- Fixes slow playback of large recording files at larger time indices
- Reduces memory use of source IDs by adding them to the common string pool
- Fixes an issue with not playing data back on the FX thread
- Fixes NPE when updating sources with no data
- Fixes updating NetworkTables source tree in playback (#152)
- Records initial conditions of all sources at the start of recording, instead of only recording values that change

Still to do:
- [x] Allow dragging of sources from source trees during playback


nightpool:
changes look good. dragging from source tree is probably out of scope for this PR, unless it's just a bugfix?

SamCarlberg:
I'm still working on this PR

SamCarlberg:
I ran into a bizarre issue a few times that I couldn't pin down. Occasionally during playback, the FX thread locks up. Looking at JVisualVM, the cause is lots of calls being made to `javafx.scene.control.TextFormatter.updateValue`. I unfortunately wasn't able to create a snapshot any time it happened so I don't have a stack trace to analyze. Since I can't track down the root cause, and since it happens so irregularly, I'm not going to address it as part of this PR since there are other things that need to be done ASAP.

PR #259, opened 2017-10-26
Use a tiled grid-rule background image on tile panes

SamCarlberg:
No longer uses the debug "gridLinesVisible" property

# Screenshots

grid-light

grid-dark


bradamiller:
I'm not sure if this is related to this PR, but after installing it there is an error in the grid resizing. If I had a default 128 pixel grid size, then place a boolean box, then change the grid spacing to 64 the boolean box is resized to an in-between size (not 64 or 128).

SamCarlberg:
That's unrelated

PR #263, opened 2017-10-28
 Add API for plugins to depend on one another

SamCarlberg:
This allows plugins to be automatically unloaded when a dependency plugin is unloaded, as well as being prevented from being loaded until all dependencies are loaded themselves.

This enforces plugins to use semantic versioning for their version strings


app/app.gradle.kts
```diff
@@ -39,6 +40,23 @@ tasks.withType<Jar> {
 }
 
 /**
+ * Copies the built jar(s) from the test_plugins project into the test resources directory.
+ */
+val copyTestPlugins = task<Copy>("copyTestPlugins") {
+    description = "Copies the built jar(s) from the test_plugins project into the test resources directory"
+    dependsOn(project("test_plugins").tasks["jar"])
+    from("""${project("test_plugins").buildDir}/libs""")
+    into("src/test/resources")
```
JLLeitschuh: Copy this into a `build` directory and add an additional test resource using the gradle API.
Then it will get removed when you run a `./gradlew clean`.

app/src/test/java/edu/wpi/first/shuffleboard/app/plugin/PluginLoaderTest.java
```diff
@@ -1,30 +1,356 @@
...
+    assertFalse(dependent.isLoaded(), "Dependent plugin is still loaded");
+  }
+
+  public static void assertInstanceOf(Object obj, Class<?>... possibleTypes) {
+    for (Class<?> possibleType : possibleTypes) {
+      if (possibleType.isInstance(obj)) {
+        return;
+      }
+    }
+    throw new AssertionError();
```
JLLeitschuh: Message?

settings.gradle
```diff
@@ -1,5 +1,6 @@
 include ":api"
 include ":app"
+include ":app:test_plugins"
```
JLLeitschuh: `test_plugins` and `integ-test` are following different naming conventions. I'd pick one and use it.

JLLeitschuh:
What happens if versions don't match up? Does this support version ranges? Eg: `[3.0.0,4.5.1)`?

SamCarlberg:
No support for version ranges. Better version specification would be nice to have, though. Right now it just accepts any version of _at least_ the specified one, eg requiring `1.2.3` will work if version `1.2.3`, `.1.12.18`, or `26.4.18` are present, but not `1.2.2`.

api/api.gradle.kts
```diff
@@ -13,6 +13,7 @@ dependencies {
     api(group = "de.codecentric.centerdevice", name = "javafxsvg", version = "1.2.1")
     api(group = "edu.wpi.first.ntcore", name = "ntcore-java", version = "4.+")
     api(group = "eu.hansolo", name = "Medusa", version = "7.9") // Note the capital 'M' -- lowercase is a much older version!
+    api(group = "com.cedarsoft.commons", name = "version", version = "8.3.1")
```
JLLeitschuh: Curious why you picked this instead of this other library.
https://github.com/zafarkhaja/jsemver

Did you see both and pick this one because it had better support?

app/test_plugins/src/main/java/edu/wpi/first/shuffleboard/testplugins/DependentPlugin.java
```diff
@@ -0,0 +1,15 @@
...
+import edu.wpi.first.shuffleboard.api.plugin.Plugin;
+
+/**
+ * A plugin that depends on another plugin.
+ */
+public class DependentPlugin extends Plugin {
+
+  public DependentPlugin() {
+    super("edu.wpi.first.shuffleboard", "Dependent Plugin", "1.0.0", "A plugin that depends on another plugin");
+    addDependency("edu.wpi.first.shuffleboard", "Basic Plugin", "1.0.0");
```
JLLeitschuh: If the plugin uses `edu.wpi.first.shuffleboard.testplugins.BasicPlugin.someAPInotInOneDotZeroDotZero()` in this constructor, won't this simply blow up with a `Method Def Not Found` exception making this whole dependency version stuff moot?
Wouldn't it be better to load this information from some resource file that the plugin stores in the JAR? That way you're not going to get classloader issues.

SamCarlberg: If the plugin depends on an API present in version `1.1.0` but not in `1.0.0`, the dependent version string should reflect that. Otherwise you're saying your plugin depends on version A when it _actually_ depends on version B.

api/api.gradle.kts
SamCarlberg: The cedarsoft one is smaller, easier to understand, and has built in support for checking stuff like `versionA >= versionB`. Also has a `VersionRange` class that might be useful for bounded dependency versions

settings.gradle
SamCarlberg: I'd prefer the underscore format. Not sure where other people lie on this

app/test_plugins/src/main/java/edu/wpi/first/shuffleboard/testplugins/DependentPlugin.java
JLLeitschuh: No, what I'm saying is that, if you have the wrong version on the classpath you're going to get exceptions when the classloader tries to load invalid classes or call methods that don't exist.

SamCarlberg: I think that would only crop up when the plugin class itself has static dependencies or dependencies in the constructor/instantiation blocks/fields.

Like this would crash if version 2 of the dependent plugin isn't loaded, because it'll try to access those classes before the plugin loader can check to see if the plugin can actually be loaded.

```java
public class BadPlugin extends Plugin {

  // crashes if version 2 of the dependent plugin hasn't already been loaded!
  private final Map<DataType, ComponentType> defaultComponents =
    ImmutableMap.of(
        SomeDataTypeFromVersion2, ...
    );

  public BadPlugin() {
    ...
    addDependency("...", "...", "2.0.0");
  }

  @Override
  public Map<DataType, ComponentType> getDefaultComponents() {
    return defaultComponents;
  }
}
```

---

I think there are two solutions here:

1. Only use stuff from other plugins in widgets, data types, or in the plugin _methods_; anywhere but fields or constructors in the plugin class. This is what we currently do in the builtin plugins
2. Store the variables in a static inner class that gets referenced in the plugin methods. The inner class is only loaded when first referenced by one of the methods, eg:

```java
public class MyPlugin extends Plugin {

  public MyPlugin() {
    ...
    addDependency("...", "...", "2.0.0");
  }

  public Map<DataType, ComponentType> getDefaultComponents() {
    return Holder.defaultComponents;
  }

  private static final class Holder {
    private static final Map<DataType, ComponentType> defaultComponents =
      ImmutableMap.of(
        SomeDataTypeFromVersion2, ...
      );
  }

}
```

Of course, there is always the assumption that the plugin developers remember to keep the dependencies that the plugin exports up-to-date with changes to the actual library dependencies. There's not much we can do there, besides maybe wrapping the `loadClass` calls in a try-catch.

JLLeitschuh: > Like this would crash if version 2 of the dependent plugin isn't loaded, because it'll try to access those classes before the plugin loader can check to see if the plugin can actually be loaded.

This is why gradle has the `afterEvaluate` block. Because these sorts of problems arise.

You probably need to expose some equivalent to `afterEvaluate` and tell plugin authors not to use dependent types until they are out of the constructor.

SamCarlberg: I think it'll be okay if we keep it as-is and warn plugin authors about this. There are already a few ways of working around it, and the builtin plugins all follow the first solution, so people copying the structure of those plugins shouldn't have any issues at all

JLLeitschuh: Another quick solution is annotation based processing.
Then you'd only have the problem if the plugin developer uses a plugin type in a `static` code block.

SamCarlberg: How so? Can you give an example?

settings.gradle
JLLeitschuh: Idk. I'd just make it consistent.

app/test_plugins/src/main/java/edu/wpi/first/shuffleboard/testplugins/DependentPlugin.java
JLLeitschuh:
```java
@Dependencies(
    values = [
        @Dependency("...", "...", "2.0.0"),
        @Dependency("...", "...", "3.0.0"),
    ]
)
class SomePluginName {
}
```

SamCarlberg: Hm. That could work.

.gitignore
```diff
@@ -1,3 +1,8 @@
+### Shuffleboard ###
+
+# Ignore the test plugins jar, force it to be rebuilt locally
+app/src/test/resources/test_plugins.jar
```
JLLeitschuh: I'd put resources that are generated into the `/build` directory instead of doing this. Then gradle clean will clean these up.

JLLeitschuh: Do something like `/build/generated-resources`

app/app.gradle.kts
```diff
@@ -39,6 +40,23 @@ tasks.withType<Jar> {
 }
 
 /**
+ * Copies the built jar(s) from the test_plugins project into the test resources directory.
+ */
+val copyTestPlugins = task<Copy>("copyTestPlugins") {
+    description = "Copies the built jar(s) from the test_plugins project into the test resources directory"
+    dependsOn(project("test_plugins").tasks["jar"])
+    from("""${project("test_plugins").buildDir}/libs""")
```
JLLeitschuh: Why is this a multiline string??

app/src/main/java/edu/wpi/first/shuffleboard/app/plugin/PluginLoader.java
```diff
@@ -196,11 +404,22 @@ private void tryRestoreSource(Widget widget, DestroyedSource destroyedSource) {
    * Unloads a plugin.
    *
    * @param plugin the plugin to unload
+   *
+   * @return true if the plugin was unloaded, false if not. This can occur if the plugin was not loaded to begin with.
    */
   @SuppressWarnings("unchecked")
-  public void unload(Plugin plugin) {
+  public boolean unload(Plugin plugin) {
```
JLLeitschuh: What if a user tries to unload a plugin that another plugin has a dependency on?

api/src/main/java/edu/wpi/first/shuffleboard/api/plugin/PluginArtifact.java
```diff
@@ -0,0 +1,90 @@
...
+  /**
+   * Creates a new plugin artifact.
+   *
+   * @param groupId the ID of the group that develops the plugin
+   * @param name    the name of the plugin
+   * @param version the version of the plugin
+   *
+   * @throws IllegalArgumentException if the version does not follow semantic versioning
+   */
+  public PluginArtifact(String groupId, String name, String version) {
```
JLLeitschuh: This is user input. I'd do null checks on all of these.

app/src/main/java/edu/wpi/first/shuffleboard/app/plugin/PluginLoader.java
SamCarlberg: All dependent plugins are unloaded first.

api/src/main/java/edu/wpi/first/shuffleboard/api/plugin/Plugin.java
```diff
@@ -64,9 +64,15 @@ protected Plugin() {
...
     this.groupId = description.group();
     this.name = description.name();
-    this.version = description.version();
-    this.description = description.summary();
-    this.artifact = new PluginArtifact(this.groupId, this.name, this.version);
+    this.version = Version.parse(description.version());
+    this.summary = description.summary();
+
+    if (groupId.contains(":")) {
+      throw new IllegalStateException("A plugin's group ID cannot contain a colon (':') character - " + groupId);
```
JLLeitschuh: `IllegalArgumentException`. This is not state.

api/src/main/java/edu/wpi/first/shuffleboard/api/plugin/Plugin.java
```diff
@@ -64,9 +64,15 @@ protected Plugin() {
...
+    }
+    if (name.contains(":")) {
+      throw new IllegalStateException("A plugin's name cannot contain a colon (':') character - " + name);
```
JLLeitschuh: `IllegalArgumentException`. This is not state.

api/src/main/java/edu/wpi/first/shuffleboard/api/plugin/Plugin.java
SamCarlberg: It's state set by the annotation. It's not `IllegalArgument` because there is no argument passed to the constructor

JLLeitschuh: The user is passing an argument into the annotation.

app/app.gradle.kts
SamCarlberg: Adding `test_plugins/build/libs` as a test resource directory avoids copying and also lets `gradlew clean` remove the JAR.

.gitignore
SamCarlberg: I just added `test_plugins/build/libs` as a test resource in the sourcesets. That seemed cleaner

JLLeitschuh: You should be able to remove this from the `.gitignore` now then right??

SamCarlberg: Good catch!

api/src/main/java/edu/wpi/first/shuffleboard/api/plugin/Plugin.java
```diff
@@ -26,21 +28,73 @@
...
+   *
+   * @throws InvalidPluginDefinitionException if no {@code @Description} annotation is present, or if the
+   */
+  protected Plugin() {
+    validatePluginClass(getClass());
+    Description description = getClass().getAnnotation(Description.class);
+
+    this.groupId = description.group();
+    this.name = description.name();
+    this.version = Version.parse(description.version());
```
JLLeitschuh: I'd turn this into a type where all this stuff is combined into one thing. Eg. `PluginIdentifier` type. Seems like you are conflating this information with the `Plugin` type.

If you create one type `PluginIdentifier` you can then put the validation onto the `create` method for that type.

api/src/main/java/edu/wpi/first/shuffleboard/api/plugin/Plugin.java
```diff
@@ -26,21 +28,73 @@
...
+    String name = description.name();
+    if (name.contains(":")) {
+      throw new InvalidPluginDefinitionException("The plugin name cannot contain a colon (:) character: " + name);
+    }
+    String version = description.version();
+    try {
+      Version.parse(version);
+    } catch (IllegalArgumentException e) {
+      throw new InvalidPluginDefinitionException("The version string does not follow semantic versioning", e);
+    }
```
JLLeitschuh: You might want to create one uber exception with all of the problems rolled up into one exception instead of getting each one in this order. This will help developers to resolve each problem without having to re-launch to see if anything else is wrong.

api/src/main/java/edu/wpi/first/shuffleboard/api/plugin/Requires.java
```diff
@@ -0,0 +1,46 @@
...
+   * number, and at least the same minor version number, will allow this plugin to be loaded. If no plugin is loaded
+   * with a compatible version, this plugin will not be allowed to be loaded.
+   *
+   * <p>For example, if the minimum compatible version is {@code "1.2.3"} and a plugin is present with version
+   * {@code "1.4.6"}, this plugin will be allowed to load since there is a compatible, albeit more recent, version of
+   * the dependency. However, if only version {@code "1.1.1"} is loaded, this plugin cannot be loaded because the loaded
+   * version of the dependency is too old. The same goes if a version {@code "2.2.3"} is loaded; it is too recent and
+   * does not have a backward-compatible API with {@code "1.2.3"}.
+   */
+  String minVersion();
```
JLLeitschuh: Maven supports version ranges. Something like `[4.0.3,4.5.3)` I think you mentioned you didn't want to do that though?

PR #266, opened 2017-10-28
Publish API and plugin jars and source and javadocs for all artifacts

### LINKED ISSUE #193: API jar needs to be published
Right now, we're only publishing the app uberjar as `edu.wpi.first.shuffleboard:Shuffleboard:<version>`

We should publish two jars: an API jar (may be an uberjar to give users access to guava etc.) and the fat app jar.

| Project | Artifact name |
|---|---|
| API | `shuffleboard-api` |
| APP | `shuffleboard-app-all` |

CC @JLLeitschuh

### LINKED ISSUE #265: Builtin plugin jars should be published
Biggest usecase is being able to create custom widgets for data types defined in another plugin; for example, a custom gyro display that rotates an image of a teams robot.

SamCarlberg:
Fixes #193
Fixes #265

Plugins are currently distributed as normal jars, (not as uberjars). A good improvement would be to bundle non-shuffleboard dependencies so third-party plugins dependent on the bulitin ones (see #263) are able to use their libraries, eg `ntcore` for a plugin that depends on the NetworkTables plugin. Unfortunately, it looks like it's difficult to exclude transitive dependencies so even if the API files are excluded, all of its dependent libraries (and their dependents, and so on) will be bundled in every plugin.

Since the builtin plugins are already distributed as part of the app, it's not as big a deal from our end. But third-party developers won't have access to `cscore` (and probably `ntcore` once that gets moved entirely to the nt plugin).

@JLLeitschuh Any thoughts on how we can bundle _only_ the dependencies declared in the plugin gradle files as part of a shadow jar?

### DIFF
--- a/plugins/cameraserver/cameraserver.gradle.kts
+++ b/plugins/cameraserver/cameraserver.gradle.kts
@@ -2,10 +2,6 @@ description = """
 The bundled CameraServer plugin. This plugin provides data sources and widgets for viewing MJPEG streams from the WPILib CameraServer.
 """.trimIndent().trim()
 
-plugins {
-    id("edu.wpi.first.wpilib.versioning.WPILibVersioningPlugin") version "1.6"
-}
-
 dependencies {
     api(group = "edu.wpi.first.cscore", name = "cscore-java", version = "+")
     implementation(group = "edu.wpi.first.cscore", name = "cscore-jni", version = "+")
--- a/plugins/cameraserver/cameraserver.gradle.kts
+++ b/plugins/cameraserver/cameraserver.gradle.kts
@@ -3,7 +3,9 @@ The bundled CameraServer plugin. This plugin provides data sources and widgets f
 """.trimIndent().trim()
 
 dependencies {
-    api(group = "edu.wpi.first.cscore", name = "cscore-java", version = "+")
-    implementation(group = "edu.wpi.first.cscore", name = "cscore-jni", version = "+")
-    implementation(group = "edu.wpi.first.wpilib", name = "opencv", version = "3.2.0")
+    compile(group = "edu.wpi.first.cscore", name = "cscore-java", version = "+")
+    compile(group = "org.opencv", name = "opencv-java", version = "3.2.0")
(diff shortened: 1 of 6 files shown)


SamCarlberg:
Addressed plugin publishing in 31b541c. @ThadHouse can you please check to make sure artifacts are being published as you'd expect

JLLeitschuh:
Confused why you stopped using the `java-library` plugin, but sure.

SamCarlberg:
For the base plugin? It was redundant because it's set in the root `plugins.gradle.kts`

plugins/cameraserver/cameraserver.gradle.kts
```diff
@@ -2,12 +2,10 @@ description = """
...
-
 dependencies {
-    api(group = "edu.wpi.first.cscore", name = "cscore-java", version = "+")
-    implementation(group = "edu.wpi.first.cscore", name = "cscore-jni", version = "+")
-    implementation(group = "edu.wpi.first.wpilib", name = "opencv", version = "3.2.0")
+    compile(group = "edu.wpi.first.cscore", name = "cscore-java", version = "+")
+    compile(group = "org.opencv", name = "opencv-java", version = "3.2.0")
+    compile(group = "org.jcodec", name = "jcodec-javase", version = "0.2.0")
+    compile(group = "edu.wpi.first.cscore", name = "cscore-jni", version = "+", classifier = "all")
+    compile(group = "org.opencv", name = "opencv-jni", version = "+", classifier = "all")
```
JLLeitschuh: I don't understand why you are using the `compile` configuration here. You shouldn't if you are using the `java-library` plugin.

ThadHouse: The JNI ones are runtime only, so these definitely should be runtime. These artifacts are huge (~100mb combined), and definitely not needed to just compile plugins.

JLLeitschuh: You should only use the `api` or `implementation` configurations.

PR #269, opened 2017-10-30
Add widget for editing robot preferences

### LINKED ISSUE #28: Toggle switch widget moves outside of slider area
When toggling a Toggle Switch widget it moves outside of the oval track on the right side. See attached screen shot.
toggleswitch

### LINKED ISSUE #175: Toggle Switches in Tile property sheets are invisible

SamCarlberg:
Move WidgetPropertySheet to API so plugins can use it
Make text fields on dark theme only have colored underline when focused
Relies on wpilibsuite/allwpilib#701 for type metadata

Fixes #28 (for real this time) by updating to a newer version of controlsfx
Fixes #175

Search field is toggleable in the widget settings

prefs-0
prefs-1
prefs-2
prefs-3

### DIFF
--- a/api/src/main/java/edu/wpi/first/shuffleboard/api/components/WidgetPropertySheet.java
+++ b/api/src/main/java/edu/wpi/first/shuffleboard/api/components/WidgetPropertySheet.java
@@ -28,7 +28,7 @@ import javafx.util.StringConverter;
 public class WidgetPropertySheet extends PropertySheet {
 
   /**
-   * Creates an empry property sheet.
+   * Creates an empty property sheet.
    */
   public WidgetPropertySheet() {
     super();
--- a/api/src/main/java/edu/wpi/first/shuffleboard/api/components/WidgetPropertySheet.java
+++ b/api/src/main/java/edu/wpi/first/shuffleboard/api/components/WidgetPropertySheet.java
@@ -1,7 +1,5 @@
-package edu.wpi.first.shuffleboard.app.components;
+package edu.wpi.first.shuffleboard.api.components;
 
-import edu.wpi.first.shuffleboard.api.components.IntegerField;
-import edu.wpi.first.shuffleboard.api.components.NumberField;
 import edu.wpi.first.shuffleboard.api.theme.Theme;
 import edu.wpi.first.shuffleboard.api.theme.Themes;
 
@@ -11,7 +9,7 @@ import org.controlsfx.property.editor.AbstractPropertyEditor;
 import org.controlsfx.property.editor.DefaultPropertyEditorFactory;
 import org.controlsfx.property.editor.PropertyEditor;
 
-import java.util.List;
+import java.util.Collection;
 import java.util.Optional;
 import java.util.stream.Collectors;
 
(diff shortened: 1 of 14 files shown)


api/src/main/java/edu/wpi/first/shuffleboard/api/components/WidgetPropertySheet.java
```diff
@@ -31,12 +28,10 @@
 public class WidgetPropertySheet extends PropertySheet {
 
   /**
-   * Creates a new property sheet for the given widget.
```
nightpool: empty

api/src/main/java/edu/wpi/first/shuffleboard/api/components/WidgetPropertySheet.java
```diff
@@ -31,12 +28,10 @@
...
+   * Creates an empry property sheet.
```
nightpool: typo

api/src/main/java/edu/wpi/first/shuffleboard/api/components/WidgetPropertySheet.java
```diff
@@ -31,12 +28,10 @@
 public class WidgetPropertySheet extends PropertySheet {
```
nightpool: name is misleading now that we're using it for things that aren't widgets.

plugins/base/src/main/java/edu/wpi/first/shuffleboard/plugin/base/widget/RobotPreferencesWidget.java
```diff
@@ -0,0 +1,90 @@
...
+        prevData.asMap().entrySet().stream()
+            .map(Map.Entry::getKey)
+            .filter(k -> !curData.containsKey(k))
+            .forEach(wrapperProperties::remove);
+      }
+      updated.forEach((key, value) -> {
+        if (NetworkTableUtils.isMetadata(key)) {
+          return;
+        }
+        if (wrapperProperties.containsKey(key)) {
```
nightpool: This can be cleaned up with computeIfAbsent

plugins/base/src/main/java/edu/wpi/first/shuffleboard/plugin/base/data/RobotPreferencesData.java
```diff
@@ -0,0 +1,21 @@
...
+import java.util.Map;
+
+public class RobotPreferencesData extends MapData {
+
+  public RobotPreferencesData(Map<String, Object> map) {
+    super(map);
+  }
+
+  @Override
+  public RobotPreferencesData put(String key, Object value) {
```
nightpool: why override this? Seems unnecessary now that the change to put was made in MapData.

nightpool: Oh, typing. Should just call super() then.

SamCarlberg: This is the correct implementation. Calling super will return a new `MapData` instance.

api/src/main/java/edu/wpi/first/shuffleboard/api/components/WidgetPropertySheet.java
SamCarlberg: Should be a separate PR. I'll open an issue for it

PR #273, opened 2017-11-04
FIx markdown style typo in readme

PeterMitrano:


PeterJohnson:
Maybe add backticks to show both commands in fixed point font too? Also would be more consistent with the other command

PR #277, opened 2017-11-07
Automatically load the most recently saved dashboard file

### LINKED ISSUE #172: automatically re-open last used layout
Dependent on #169 

Beta if we can get to it.

### LINKED ISSUE #281: Sources in a save file that are not present at load are not later set when present
Basically if you save a file, then open it later with different/no sources, the widgets in the dashboard have no sources even when they're made available again. This makes it impossible to, for example, load a layout for playback before loading the recording file, or load a save file before starting an FRC match

SamCarlberg:
This behavior can be toggled on or off in the application settings
Refactors DashboardTab into its own upper-level class
Make widgets disable themselves, not the tile, when a source is not connected

Closes #172
Fixes #281 (since 9e9c526)

### DIFF
--- a/app/src/main/java/edu/wpi/first/shuffleboard/app/MainWindowController.java
+++ b/app/src/main/java/edu/wpi/first/shuffleboard/app/MainWindowController.java
@@ -16,6 +16,7 @@ import edu.wpi.first.shuffleboard.api.util.FxUtils;
 import edu.wpi.first.shuffleboard.api.util.Storage;
 import edu.wpi.first.shuffleboard.api.util.TypeUtils;
 import edu.wpi.first.shuffleboard.api.widget.Components;
+import edu.wpi.first.shuffleboard.app.components.DashboardTab;
 import edu.wpi.first.shuffleboard.app.components.DashboardTabPane;
 import edu.wpi.first.shuffleboard.app.components.WidgetGallery;
 import edu.wpi.first.shuffleboard.app.components.WidgetPropertySheet;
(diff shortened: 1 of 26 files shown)


app/src/main/java/edu/wpi/first/shuffleboard/app/MainWindowController.java
```diff
@@ -75,6 +76,7 @@
 /**
  * Controller for the main UI window.
  */
+@SuppressWarnings("PMD.GodClass") // TODO refactor this class
```
JLLeitschuh: 🤦‍♂️

SamCarlberg: I'm honestly surprised this PMD rule got triggered since the diff is so minor

JLLeitschuh: https://github.com/search?utf8=%E2%9C%93&q=%40SuppressWarnings%28%22PMD.GodClass%22%29+extension%3Ajava&type=Code

SamCarlberg: Going to do that refactor after the currently open PRs are merged in. Don't want to force merge conflicts if I can avoid it

app/src/main/java/edu/wpi/first/shuffleboard/app/json/SourcedRestorer.java
```diff
@@ -0,0 +1,57 @@
...
+   * Attempts to restore all destroyed sources in a sourced object for the given URIs.
+   *
+   * @param sourced       the sourced to restore sources for
+   * @param urisToRestore the URIs of the sources to restore
+   * @param errorHandler  a function to call when a source could not be restored, or when a restored source could not
+   *                      be added
+   */
+  public void restoreSourcesFor(Sourced sourced,
+                                Collection<? extends String> urisToRestore,
+                                BiConsumer<DestroyedSource, ? super Throwable> errorHandler) {
```
SamCarlberg: An error handler gets passed in so that sources can still be restored if an exception was thrown in the restoration loop. Right now all this does is log the error and say which source wasn't able to be restored.

PR #286, opened 2017-11-10
Right-click actions for layouts

nightpool:
Adds a new "ActionList" types, registers suppliers for ActionLists in the scene graph, and fetches them when creating context menu items.

image

image

Adds actions for both SubsystemLayouts and ListLayouts.


app/src/main/java/edu/wpi/first/shuffleboard/app/components/WidgetPane.java
```diff
@@ -100,6 +100,7 @@ public WidgetPane() {
       if (highlight) {
         getChildren().add(gridHighlight);
         gridHighlight.toFront();
+        gridHighlight.setMouseTransparent(true);
```
nightpool: fixes drag-and-drop regression from #259

app/src/main/java/edu/wpi/first/shuffleboard/app/components/Tile.java
```diff
@@ -34,6 +35,8 @@
   private final Property<TileSize> size = new SimpleObjectProperty<>(this, "size", null);
   private final BooleanProperty selected = new PseudoClassProperty(this, "selected");
 
+  private final PropertyBinding<String> titleProperty;
```
nightpool: fixes title bindings getting GC'd

app/src/main/java/edu/wpi/first/shuffleboard/app/components/Tile.java
```diff
@@ -49,14 +52,19 @@ protected Tile() {
...
+    titleProperty = EasyBind.monadic(contentProperty()).selectProperty(Component::titleProperty);
+    ((EditableLabel) lookup("#titleLabel")).textProperty().bindBidirectional(titleProperty);
     ((Label) lookup("#titleLabel").lookup(".label")).setTextOverrun(OverrunStyle.LEADING_ELLIPSIS);
     contentView.addListener((__, oldContent, newContent) -> {
       getContentPane()
           .map(Pane::getChildren)
-          .ifPresent(c -> c.setAll(newContent));
+          .ifPresent(c -> {
+            if (newContent != null) {
+              c.setAll(newContent);
```
nightpool: fixes NPE when wrapping widgets in layout.

plugins/base/src/main/java/edu/wpi/first/shuffleboard/plugin/base/layout/SubsystemLayout.java
```diff
@@ -51,21 +54,50 @@
...
+      }
+      return al;
+    });
   }
 
+  private WeakHashMap<Component, Pane> panes = new WeakHashMap<>();
+
   private Pane paneFor(Component component) {
+    if (panes.containsKey(component)) {
+      return panes.get(component);
```
nightpool: this pattern (EasyBind.map -> WeakHashMap -> func) appears a couple of places in the code, maybe should extract a util?

app/src/main/java/edu/wpi/first/shuffleboard/app/components/Tile.java
SamCarlberg: Already fixed in #206

app/src/main/java/edu/wpi/first/shuffleboard/app/components/Tile.java
SamCarlberg: Already fixed in #279

app/src/main/java/edu/wpi/first/shuffleboard/app/components/WidgetPane.java
SamCarlberg: So _that's_ what that was! It's been driving me nuts.

app/src/main/java/edu/wpi/first/shuffleboard/app/components/Tile.java
nightpool: I thought I remember seeing this.... lol, I thought I pulled before making this PR, must have forgotten.

nightpool:
I'm fixing the build errors/conflicts currently, feel free to review while I take care of that....

api/src/main/java/edu/wpi/first/shuffleboard/api/components/ActionList.java
```diff
@@ -0,0 +1,90 @@
...
+  private final String name;
+  private final List<Supplier<MenuItem>> actions;
+
+  private ActionList(String name) {
+    this.name = name;
+    this.actions = new ArrayList<>();
+  }
+
+  public ActionList addAction(String name, Runnable r) {
+    actions.add(() -> FxUtils.menuItem(name, _e -> r.run()));
```
SamCarlberg: I typically use a double underscore `__` for unused lambda parameters. Easier to read IMO

api/src/main/java/edu/wpi/first/shuffleboard/api/components/ActionList.java
```diff
@@ -0,0 +1,90 @@
...
+import java.util.List;
+import java.util.Optional;
+import java.util.function.Supplier;
+import java.util.stream.Collectors;
+
+import javafx.scene.Node;
+import javafx.scene.control.Menu;
+import javafx.scene.control.MenuItem;
+
+public class ActionList {
```
SamCarlberg: What is this class? What does it do? How is it intended to be used? Documentation, please!

plugins/base/src/main/java/edu/wpi/first/shuffleboard/plugin/base/layout/SubsystemLayout.java
SamCarlberg: SubsystemLayout and ListLayout should really inherit from a common superclass that keeps all this code in the same place.

plugins/base/src/main/java/edu/wpi/first/shuffleboard/plugin/base/layout/SubsystemLayout.java
```diff
@@ -51,21 +54,50 @@
   private Subscription retained; //NOPMD field due to GC
   private DataSource<?> source;
 
+  private List<Component> hidden = new ArrayList<>();
```
SamCarlberg: Why does only `SubsystemLayout` get this behavior? I'd expect to be able to hide widgets in _any_ type of layout

plugins/base/src/main/java/edu/wpi/first/shuffleboard/plugin/base/layout/ListLayout.java
```diff
@@ -42,16 +44,58 @@ private void initialize() {
...
+        widgets.add(0, component);
+      });
+    }
+
+    if (i < widgets.size() - 1) {
+      al.addAction("Move down", () -> {
+        widgets.remove(i);
+        widgets.add(i + 1, component);
+      });
+      al.addAction("Send to bottom", () -> {
```
SamCarlberg: These should also belong to `SubsytemLayout`. You'll probably have to extract a common super class, since the only real differences are

1. SubsytemLayout sorts widgets by title, alphabetically
2. SubsystemLayout has a data source

And probably an abstract superclass of _that_ for taking care of creating panes etc. for widgets, since that code is very likely to be the same for all or most layouts

plugins/base/src/main/java/edu/wpi/first/shuffleboard/plugin/base/layout/ListLayout.java
```diff
@@ -42,16 +44,58 @@ private void initialize() {
...
+    panes.put(component, pane);
     return pane;
   }
 
+  private ActionList actionsForComponent(Component component) {
+    ActionList al = ActionList.withName(component.getTitle());
+    int i = widgets.indexOf(component);
+
+    if (i > 0) {
+      al.addAction("Move up", () -> {
```
SamCarlberg: 👍

plugins/base/src/main/java/edu/wpi/first/shuffleboard/plugin/base/layout/ListLayout.java
nightpool: Why? If SubsystemLayout sorts things by title, isn't it unwanted for users to be able to reorder widgets themselves?

SamCarlberg:
I've noticed it's very hard to tell the difference between widgets and the background. A little CSS on `.layout-stack` could add backgrounds and borders to the wrapper panes. Some (configurable) spacing would be good to have, too.

Layouts should also have an "Unwrap" action to pull all their components into that layout's parent, then delete the layout.

plugins/base/src/main/java/edu/wpi/first/shuffleboard/plugin/base/layout/ListLayout.java
SamCarlberg: The layout doesn't keep the children sorted. It just kind of assumes they're sorted when deciding where to place a new child. Alphabetical sorting isn't likely to put components in a _logical_ ordering, which is why users should be able to customize the order

nightpool:
fixed:

 - code changes
 - documentation
 - subsystemlayout inherits from listlayout

not fixed:

 - unwrap action
 - show as.... action
 - edit properties

Those should be pretty simple to add on as followups though

>Context menus on a widget tile shouldn't have the "title" menu at the top; it's redundant.

Not sure what you mean by this?

PR #290, opened 2017-11-14
Use source IDs when autopopulating instead of sources

### LINKED ISSUE #285: Autopopulator creates new sources instead of checking by source ID
This causes CameraServer sources to be opened and start streaming, even if no widgets are created, resulting in unwanted bandwidth use.

Also, creating sources before knowing if they're usable is just bad practice

SamCarlberg:
Fixes #285


JLLeitschuh:
rebase on master?

PR #292, opened 2017-11-16
Make loading plugins and widgets more error-resistant

SamCarlberg:
- `ComponentType.get()` can throw an exception if the component could not be instantiated for any reason
- Fixed data types returning an unrelated type if no registered data types are in the type hierarchy
- Enforce `SingleTypeWidget` having only a single data type

Hopefully, this should make it easier to catch errors when working on adding widgets or data types. This also makes the UI not break when an exception is thrown when attempting to add a widget to the gallery or WidgetPane


app/src/main/java/edu/wpi/first/shuffleboard/app/json/WidgetSaver.java
```diff
@@ -35,8 +36,13 @@ public JsonElement serialize(Widget src, JsonSerializationContext context) {
...
     JsonObject obj = json.getAsJsonObject();
     String type = obj.get("_type").getAsString();
-    Widget widget = Components.getDefault().createWidget(type)
-            .orElseThrow(() -> new JsonParseException("No widget found for " + type));
+    Widget widget;
+    try {
+      widget = Components.getDefault().createWidget(type)
+          .orElseThrow(() -> new JsonParseException("No widget found for " + type));
+    } catch (ComponentInstantiationException e) {
+      throw new JsonParseException(e);
```
JLLeitschuh: Add another message to assist with debugging.

SamCarlberg:
Stale

PR #297, opened 2017-11-18
Fix subsystems populating with treetables for unknown complex data types

SamCarlberg:


bradamiller:
Do you have a sample OutlineViewer save file that demonstrates the issue and can be used to verify the fix?

plugins/base/src/main/java/edu/wpi/first/shuffleboard/plugin/base/layout/SubsystemLayout.java
```diff
@@ -109,9 +111,13 @@ public String getName() {
 
   @Override
   public boolean supports(String sourceId) {
+    DataType<?> dataType = SourceTypes.getDefault()
+        .typeForUri(sourceId)
+        .dataTypeForSource(DataTypes.getDefault(), sourceId);
     return getSource() != null
         && !getSource().getId().equals(sourceId)
         && sourceId.startsWith(getSource().getId())
+        && dataType != DataTypes.Map
```
JLLeitschuh: Reference equality vs. `.equals`?

PR #298, opened 2017-11-19
Save component titles

SamCarlberg:
Widget titles are only generated when modifying the sources and when the title hasn't been externally set by a user or enclosing layout

Fixes [artf5585](https://usfirst.collab.net/sf/go/artf5584)


sciencewhiz:
This works for me

bradamiller:
After applying this fix I still can't edit the title of widgets dragged from the palette (the problem described in the referenced PR). I was testing by double-clicking on the title bar. Is that what it's supposed to fix?

SamCarlberg:
No. This _only_ fixes the titles not being saved.

PR #303, opened 2017-11-20
Save tile size, hgap, and vgap

### LINKED ISSUE #209: Saving the layout doesn't save the grid size
Reloading a layout after changing the grid size doesn't work correctly. I believe the grid size is not saved.

SamCarlberg:
Breaks older save files
Fixes #209


SamCarlberg:

app/src/test/java/edu/wpi/first/shuffleboard/app/json/DashboardTabPaneSaverTest.java
```diff
@@ -1,66 +1,133 @@
...
+    components.register(new DummyBooleanBoxType());
   }
 
-  @AfterEach
-  public void tearDown() {
-    AsyncUtils.setAsyncRunner(FxUtils::runOnFxThread);
-    NetworkTableUtils.shutdown();
+  @AfterAll
+  public static void tearDown() {
+    Components.setDefault(new Components());
```
JLLeitschuh: Generally, you shouldn't need to do this sort of thing in well decoupled software.

SamCarlberg:
@bradamiller That's because the minimum tile size is 64x64. There's a bug (fixed in #306) that doesn't validate the user input so you're allowed to set the size below that, but the size is checked when it's set via `setTileSize`. The housekeeping PR also lowers the minimum tile size to 32x32.

PR #306, opened 2017-11-21
General housekeeping cleanup

### LINKED ISSUE #44: Do not use the SmartDashboard directory
People may want to still use SmartDashboard.  Are we sure what we put in the directory will not conflict?

### LINKED ISSUE #115: Widgets dragged from the gallery don't highlight the widget pane

### LINKED ISSUE #127: Changing Tile Size does not update Green outline when dragging
Please view this video:
ezgif-3-8310e7243d

### LINKED ISSUE #142: improve WidgetPropertySheet documentation
In particular this class has poor documentation:

https://github.com/wpilibsuite/shuffleboard/blob/master/app/src/main/java/edu/wpi/first/shuffleboard/app/widget/WidgetPropertySheet.java#L159-L162

Why is their implementation "bad"?

### LINKED ISSUE #221: The label for a widget inside a layout doesn't show with the ... on the left side
The names of widgets placed directly in the grid show the rightmost part of the name if it is too big to fit. The names of widgets inside a list don't do that.

### LINKED ISSUE #264: Widget class names should all end with "widget"
This would fix some headaches when trying to use UI controls that have the same or similar name to a widget

| Widget | Confused With | New Name |
|---|---|---|
| `NumberSlider` | `javafx.scene.control.Slider` | `NumberSliderWidget` |
| `ToggleButton` | `javafx.scene.control.ToggleButton` | `ToggleButtonWidget` |
| `ToggleSwitch` | `org.controlsfx.control.ToggleSwitch` | `ToggleSwitchWidget` |
| `TextView` | `TextField`, `TextArea` | `TextViewWidget` |
| `BooleanBox` | N/A | `BooleanBoxWidget` |
| `SpeedController` | N/A | `SpeedControllerWidget` |

### LINKED ISSUE #272: WidgetPropertySheet is poorly named
It's not explicitly for widgets, and is used mostly to have better editors for numbers and booleans

SamCarlberg:
Fixes lots of minor issues:

- Fixes #44
- Fixes #115
- Fixes #127
- Fixes #142
- Fixes #221
- Fixes #264
- Fixes #272

Also improves performance of previewing tiles from dropped sources by caching the size rather than creating and computing the size of a new widget/layout EVERY time the mouse is moved while dragging


nightpool:
What's going on with SendableChooserAdapter here? seems to have disappeared?

SamCarlberg:
It was unused so it was removed. Serializers for FRC data types should be all-or-nothing; having just one adapter for one data type out of many just isn't good. It implies that we support serializing FRC data not from NetworkTables, which we don't

api/src/main/java/edu/wpi/first/shuffleboard/api/data/DataTypes.java
```diff
@@ -24,17 +33,91 @@
...
+/**
+ * Registry of data types in shuffleboard. This class also provides data types for various "wildcard" types, as well as
+ * types for primitive and arrays of primitive data. These types are always registered and may not be unregistered.
+ */
 public class DataTypes extends Registry<DataType> {
 
   // TODO replace with DI eg Guice
   private static DataTypes defaultInstance = null;
 
   // Catchall or wildcard types
```
JLLeitschuh: Catch-all

api/src/main/java/edu/wpi/first/shuffleboard/api/data/DataTypes.java
```diff
@@ -24,17 +33,91 @@
...
+  /**
+   * The type corresponding to an array of raw bytes (<tt>byte[]</tt>).
+   */
+  public static final DataType<byte[]> ByteArray = new RawByteType();
+
+  /**
+   * The default data types. None of these may be unregistered.
+   */
+  private static final ImmutableCollection<DataType<?>> defaultTypes = ImmutableSet.of(
+      // catchall
```
JLLeitschuh: catch-all

api/src/main/java/edu/wpi/first/shuffleboard/api/sources/recording/serialization/Serializers.java
```diff
@@ -17,6 +19,16 @@
 
   static {
     Runtime.getRuntime().addShutdownHook(new Thread(Serializers::cleanUpAll));
```
JLLeitschuh: At a high level, can you explain why this is necessary?

app/src/main/java/edu/wpi/first/shuffleboard/app/components/TilePane.java
```diff
@@ -409,6 +409,9 @@ public boolean isOverlapping(int col, int row, int tileWidth, int tileHeight, Pr
     return false;
   }
 
+  /**
+   * Gets the layout of a tile in this tile pane.
```
JLLeitschuh: Gets or creates??

app/src/main/java/edu/wpi/first/shuffleboard/app/dnd/ResizeUtils.java
```diff
@@ -0,0 +1,37 @@
...
+import javafx.scene.Node;
+
+/**
+ * A utility class for dealing with resizing tiles.
+ */
+public final class ResizeUtils {
+
+  public static final Predicate<Node> isResizedTile = n -> n instanceof Tile && isResizedTile((Tile) n);
+
+  private static final AtomicReference<Tile<?>> currentTile = new AtomicReference<>(null);
```
JLLeitschuh: This seems a bit concerning. Why is this necessary?

app/src/main/java/edu/wpi/first/shuffleboard/app/components/TilePane.java
SamCarlberg: It computes the layout and stores it in a data object? It only "creates" a layout in the very narrow sense that a new data object is created to contain it

api/src/main/java/edu/wpi/first/shuffleboard/api/sources/recording/serialization/Serializers.java
SamCarlberg: Cleans up serializers eg by closing file handles or finalizing recorded video files. Otherwise recorded video files would be unreadable

plugins/base/src/test/java/edu/wpi/first/shuffleboard/plugin/base/recording/serialization/SendableChooserAdapterTest.java
```diff
@@ -1,37 +0,0 @@
...
-
-import edu.wpi.first.shuffleboard.plugin.base.data.SendableChooserData;
-
-import org.junit.jupiter.api.Test;
-
-import static org.junit.jupiter.api.Assertions.assertArrayEquals;
-import static org.junit.jupiter.api.Assertions.assertEquals;
-
-
-public class SendableChooserAdapterTest extends AbstractAdapterTest<SendableChooserData> {
```
JLLeitschuh: Are you not supporting SenableChooser's anymore?

api/src/main/java/edu/wpi/first/shuffleboard/api/sources/recording/serialization/Serializers.java
JLLeitschuh: Fair enough. So this only works if the app fails "gracefully" then (instead of via force-quit).

plugins/base/src/test/java/edu/wpi/first/shuffleboard/plugin/base/recording/serialization/SendableChooserAdapterTest.java
SamCarlberg: All data recording for FRC data types is currently done by the NetworkTableSourceType to reduce file size. See https://github.com/wpilibsuite/shuffleboard/pull/306#issuecomment-346453399

api/src/main/java/edu/wpi/first/shuffleboard/api/data/DataTypes.java
SamCarlberg: That's just an alternate spelling.

app/src/main/java/edu/wpi/first/shuffleboard/app/dnd/ResizeUtils.java
SamCarlberg: It's so that the widget pane can ignore the tile being resized when previewing its layout. Otherwise the highlight would always be red/"invalid" instead of only when it intersects with another tile

bradamiller:
has merge conflicts

app/src/main/java/edu/wpi/first/shuffleboard/app/dnd/ResizeUtils.java
JLLeitschuh: Not a fan of this but I don't have a better solution that doesn't involve some sort of DI framework.

Eg. The GRIP Drag service: https://github.com/WPIRoboticsProjects/GRIP/blob/1c1a5f6faa310dfd82b576632c46f2e9e00dc728/ui/src/main/java/edu/wpi/grip/ui/dragging/DragService.java

Issue #293, opened 2017-11-17
Add warning to edit grade-wrapper.properties in README


aswerdlow935:
Tried to shuffleboard with both Java JDK 8 & 9 installed (at least on a mac), and shuffleboard defaulted to Java JDK 9 even after gradle.property and bash_profile changes. However, adding `javaHome=/Library/Java/JavaVirtualMachines/jdk1.8.0_131.jdk/Contents/Home` fixes the issue. A warning should be added in the README to add this property if running fails with the unknown 9.0.1 error message.

JLLeitschuh:
More robust solution, before you run gradle, you just run this:
```bash
JAVA_HOME=`/usr/libexec/java_home -v 1.8`
```
or if you are using the Fish shell:
setenv -gx JAVA_HOME (/usr/libexec/java_home -v 1.8)

aswerdlow935:
That also works just as well...thanks👌

PR #308, opened 2017-11-22
README: How to set JAVA_HOME if JDK 9 is installed

JLLeitschuh:
Related: #293

### DIFF
--- a/README.md
+++ b/README.md
@@ -17,6 +17,17 @@ To run shuffleboard use the command `./gradlew :app:run`.
 ### Requirements
 - [JDK 8](http://www.oracle.com/technetwork/java/javase/downloads/index.html)
 
+If you have both the JDK 9 and JDK 8 installed on your machine you may need to explicitly set your shell's enviroment variables before running Gradle.
+
+For Bash:
+```bash
+JAVA_HOME=`/usr/libexec/java_home -v 1.8`
+```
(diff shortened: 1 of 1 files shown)


SamCarlberg:
What about Windows?

JLLeitschuh:
I think you have to do some wackiness with the environment variables through the UI.

When I was working on WPILib I used this tool which was 100% better than the one that was built into Windows 7.
http://eveditor.com/

SamCarlberg:
FWIW Windows 10 has a _much_ better environment variable editor then previous versions of Windows. 

It may be possible to use `set` to change `PATH` an `JAVA_HOME` in a local command session without affecting the actual environment variables. It'd be tricky, though. Ugh.

sciencewhiz:
does java -Djava.home=PATH work? That would be cross platform

JLLeitschuh:
No, that wouldn't work. You need to have `$JAVA_HOME` set for gradle to be executed with the right version. Look at the code inside of the `gradlew` file.

JLLeitschuh:
Anyone want to merge this??

JLLeitschuh:
@SamCarlberg Do you want me to close this?

SamCarlberg:
I think so, now that we display a warning dialog and prevent launching when not running on Java 8

PR #315, opened 2017-12-14
Add widgets for mecanum and differential drive bases

SamCarlberg:
Differential drive displays an arrow showing the motion of the drive base. The arrow curves to match the arc in which the robot travels, and its length matches the speed of the robot.

Currently a WIP due to the mecanum widget requiring curved arrows and some hysteresis needed on the differential drive widget to avoid repeated (and distracting) arrow flips when the sides are near-equivalent speeds


bradamiller:
I like the differential drive widget, but it is pretty huge for a small screen. It looks like its about 320 pixels wide at it's minimum size. So on a 1024x768 screen, it takes up a third of the width of the screen. I like the arrow, but it does some weird stuff at the edges of the range, that I noticed when the two sides were going almost full speed in opposite directions.

PR #319, opened 2017-12-24
Fix bug with camera stream images

SamCarlberg:
Images would have a strip of pixels from the right edge shown on the left side of the image
Also fixes the video sink not being connected when the stream URLs are set after a source is created
Fixes CameraServer widget not reconnecting when losing and regaining connection to the server


sciencewhiz:
This reconnects the camera for me now.

PR #324, opened 2018-01-02
Show connection indicators for sources

### LINKED ISSUE #54: Add NetworkTable connection indicator

SamCarlberg:
Each source type has its own indicator
Source types may opt out with the `@UiHints` annotation
Supersedes #232, closes #54


bradamiller:
Doesn't work if the connection goes away.

SamCarlberg:
That looks like an ntcore issue, where it isn't posing a connection notification when disconnecting from a robot's network

PR #348, opened 2018-01-05
Make supported java versions more clear in readme

sciencewhiz:


README.md
```diff
@@ -15,7 +15,7 @@ subprojects that the main app depends on to provide data types, widgets, and dat
 To run shuffleboard use the command `./gradlew :app:run`.
```
SamCarlberg: Since gradle requires a JDK, maybe update this line to say something like "To run shuffleboard from source, use..." and then a little bit more info for running the jar, eg clicking the jar in a file viewer or through the Eclipse plugins.

sciencewhiz: Moved that part to the building section and mentioned eclipse plugins in the running section.

PR #369, opened 2018-01-17
Rewrite of PluginLoader getRequirements to fix Eclipse error

rlee287:
With the `Stream.concat` method, Eclipse would complain that the return type should be `List<Object>`. This is a slight adjustment to the code to avoid this issue by combining the contents separately. I locally ran the `./gradlew spotlessCheck check jacocoJunit5TestReport --stacktrace --console=plain -PlogTests -Pheadless` command from `.travis.ubuntu.sh` on my computer (Ubuntu 17.10) to confirm that the tests pass.

### DIFF
--- a/README.md
+++ b/README.md
@@ -4,7 +4,6 @@
 
 # shuffleboard
 
-This is the 604 fork of shuffeboard. **The `604version` branch will be repeatedly forced-pushed as necessary in order to bring in updates from the main repo.**
 
 ## Structure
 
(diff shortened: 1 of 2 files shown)


SamCarlberg:
I appreciate the contribution, but this is more a problem with Eclipse's compiler and autodetection. You can either ignore it in Eclipse and build with gradle, since that will use javac from your JDK instead of the Eclipse compiler, or you can use an IDE with better autodetection (IntelliJ, Netbeans).

Issue #370, opened 2018-01-18
Turning on 'Auto populate' before saving Source prefix behaves unpredictably


garrett-davidson:
Tab auto-population does not work as expected. After much testing, here are my current findings:

All of this was run with the same C++ robot code. We have one axis camera added to the camera server, the default Live Window setup, and several items of various types being sent via SmartDashboard with various prefixes. The testing was centered around our SmartDashboard values with the "Auton" prefix (e.g. "Auton/lTurn1").

Test 1:
Create a tab.
Open tab preferences.
Set prefix to "SmartDashboard/Auton/".
Switch on Auto populate.
Click close.

Result: Every item in NetworkTable, including all of our camera settings, is displayed with fully qualified names (e.g. "CameraPublisher/Axis Camera/Property/exposure_priority", "SmartDashboard/LeftDriveEncoder/Speed"). However, everything with the "Auton" prefix had the prefix stripped (e.g. "lTurn1").

Test 2:
Create a tab.
Open tab preferences.
Set prefix to "SmartDashboard/Auton" (no trailing /).
Switch on Auto populate.
Click close.

Result: Same as Test 1, but several of the widgets changed order, and the "Auton" prefix values no longer had the prefix stripped (e.g. "SmartDashboard/Auton/lTurn1").

Test 3: *Finally figured out how to make it work correctly*
Create a tab.
Open tab preferences.
Set prefix to "SmartDashboard/Auton" (no trailing /).
**Click close.**
**Open tab preferences.**
Switch on Auto populate.
Click close.

Result: Everything with our "Auton" prefix showed up correctly, with the prefix properly stripped, and nothing else appeared.

SamCarlberg:
Duplicate of #354 
Closed by #355 

If you're running shuffleboard from the WPILib Eclipse plugins, the fix will be distributed in the next plugins release. That's currently slated for tomorrow night

PR #371, opened 2018-01-18
Allow checking for and applying updates from the FRC maven server

SamCarlberg:
Shuffleboard will check for updates at startup by parsing `http://first.wpi.edu/FRC/roborio/maven/release/edu/wpi/first/shuffleboard/app/maven-metadata.xml` for the newest version, then prompt the user to download it. The jar is downloaded to a temp file, then copied and overwrites the original jar file after shutdown, then finally restarted.

This PR allows shuffleboard to update at its own pace without being tied to the update schedule of the Eclipse plugins.

Users can disable the check at startup in the application settings, and can check for updates at any time with the menu item `File -> Check for updates` or with Ctrl+U  or ⌘+U


SamCarlberg:
@AustinShalit Here's the reasoning behind this PR:

- Shuffleboard is still new, and new features and bugfixes are likely to get merged
- Only bundling shuffleboard releases in the Eclipse plugins means that releases are limited, which means that bugfixes and new features don't get distributed quickly (see #370, which had a fix merged a _week_ prior, but hasn't been released yet since there hasn't been a plugins release)
- The Eclipse plugins will always have the most up-to-date version at the time they're created. Shuffleboard may have had updates since then, which users would not be able to use without
  - manually checking the Maven server; or
  - opening an issue for a bugfix or feature request that has already been added in a more recent release, then having one of us point them to the new release
- Shuffleboard will not automatically update; that is _always_ up to the user. All it does is notify the user that a newer version is available, and provides an easy method for download and installation.
- Users don't have to keep track of anything special for shuffleboard. There is no standalone installer like NI or CTRE. As I said above, it's all in one jar file that comes with the WPILib Eclipse plugins.

I don't see any fracturing here. The only potential issue I can see is that having a specific version of the plugins does not guarantee a specific version of shuffleboard. I don't think that's a problem, since a new install will always have the same version, and the shuffleboard version is very easily checked

bradamiller:
It's important that the version number always increments. Also, are you covering the case where someone installs a newer version of shuffleboard, then installs an older version of the plugins? I assume that shuffleboard will get downgraded with the plugins install, then they'll start getting messages again about upgrading it.

SamCarlberg:
There's nothing I can do here to avoid the plugins from nuking the wpilib directory. Shuffleboard would be downgraded but would alert the user at startup (if enabled; they can always check manually with the menu).

The version number will never decrement, otherwise it breaks semantic versioning.

SamCarlberg:
# Screenshots

## Update prompt
screenshot from 2018-01-23 10-59-45

## Download progress
screenshot from 2018-01-23 11-00-09

## Restart prompt
screenshot from 2018-01-23 11-00-21

## "About" dialog

screenshot from 2018-01-23 11-15-49

Contains

- Build version and build date
- JRE version and vendor
- Operating system name and version
- "Copy to clipboard" button for the technical information

bot190:
There seems to be an issue with this feature. If you launch shuffleboard without an internet connection it complains that it is unable to connect to the update server. The modal that this error appears in doesn't have a way of closing it. This makes shuffleboard impossible to use without an internet connection. Something that seems extremely likely when connecting to robot wfi networks.
shuffleboardrequiresinternet

PR #394, opened 2018-02-01
Add support for FMS info

SamCarlberg:
Adds a widget for displaying game-agnostic FMS information.
Adds a plugin for displaying the Power Up field configuration.
Adds a utility class to the API for working with bitfields and flags.

# Screenshot

Top widget is the game-agnostic widget. This also shows a warning at the bottom if the robot is E-stopped
Left widget is the Power Up field
Right widget is the field without any data

image


JLLeitschuh:
Shouldn't this be in a plugin specific to this years game?

AustinShalit:
@JLLeitschuh  It is?

Issue #397, opened 2018-02-02
Refuses to Open


GaganBhat:
Hi, Shuffleboard used to work for me around a month ago, but recently it seems to just not open up at all. I am running Java version 8 update 162, so not sure if that changes anything at all. Opening through gradleRIO causes the crash as well.

sciencewhiz:
What was the resolution?

GaganBhat:
There wasn't any, the JAR still refuses to open, but I can now open it through GradleRio, so that's a good thing I guess. I wanted to collect more data before I opened an issue actively.

AustinShalit:
What operating system are you on?

GaganBhat:
Windows 10 Pro 64 Bit.

PR #399, opened 2018-02-04
Show alert when on java 9

AustinShalit:
Possibly "solves" #397

### DIFF
--- a/app/src/main/java/edu/wpi/first/shuffleboard/app/Shuffleboard.java
+++ b/app/src/main/java/edu/wpi/first/shuffleboard/app/Shuffleboard.java
@@ -74,6 +74,21 @@ public class Shuffleboard extends Application {
     Thread.currentThread().setUncaughtExceptionHandler(Shuffleboard::uncaughtException);
     onOtherAppStart = () -> Platform.runLater(primaryStage::toFront);
 
+    // Before we load components that only work with Java 8, check to make sure
+    // the application is running on Java 8. If we are running on an invalid
+    // version, show an alert and exit before we get into trouble.
+    if (!"1.8".equals(System.getProperty("java.specification.version"))) {
+      Alert invalidVersionAlert = new Alert(Alert.AlertType.ERROR);
+      invalidVersionAlert.setHeaderText("Invalid JRE Version!");
+      invalidVersionAlert.setContentText(
+          String.format("You are using an unsupported Java version: %s!  "
+                  + "Please download Java 8.",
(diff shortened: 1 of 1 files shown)


app/src/main/java/edu/wpi/first/shuffleboard/app/Shuffleboard.java
```diff
@@ -74,6 +75,27 @@ public void start(Stage primaryStage) throws IOException {
...
+    // version, show an alert and exit before we get into trouble.
+    String javaSpec = System.getProperty("java.specification.version");
+    String javaVersion = System.getProperty("java.version");
+    if (!"1.8".equals(javaSpec)) {
+      System.out.println(javaSpec);
+      Alert invalidVersionAlert = new Alert(Alert.AlertType.ERROR);
+      invalidVersionAlert.setHeaderText("Invalid JRE Version!");
+      invalidVersionAlert.setContentText(
+          String.format("You are using an unsupported Java version: %s\n"
+                  + "Please download Java 8 and uninstall Java %s.",
```
JLLeitschuh: Do they really have to uninstall their other version of java? They just need to make sure that `JAVA_HOME` is set correctly right?

AustinShalit: Certain jar launchers do not respect the java home environment variable.

JLLeitschuh: Then ship this with the Gradle Application plugin shell/command line launcher that will always respect `JAVA_HOME`.

PeterJohnson: Not an option for this year, as the DS runs the jar directly.

JLLeitschuh: So have the DS code search for java 8 and try to launch with java 8.

AustinShalit: Due to the approaching deadline of today, I think we go with this as is and circle back after season is over to make shuffleboard work with both JRE 8 & 9 or make this a better UX.

AustinShalit:
https://github.com/wpilibsuite/shuffleboard/pull/399#discussion_r166940473

> Due to the approaching deadline of today, I think we go with this as is and circle back after season is over to make shuffleboard work with both JRE 8 & 9 or make this a better UX.

PR #411, opened 2018-02-13
Fix interdependent custom plugins

### LINKED ISSUE #410: A custom plugin that depends on code in another custom plugin cannot be loaded
Specifically, if Plugin A resides in one JAR and Plugin B resides in a separate JAR, and Plugin B uses classes defined in the Plugin A JAR, then Plugin B can never be loaded.

This is due to the current plugin loading mechanism that uses one classloader per plugin JAR, so the classloader for Plugin B has no knowledge of the classes from Plugin A.

For example, this fails if `SomeClass` is in a different JAR than `PluginB`:

```java
package plugin.a;

public class SomeClass {
  public static void someMethod() {
    // ...
  }
}
```
```java
package plugin.b;

import plugin.a.SomeClass;

public class PluginB {
  @Override
  public void onLoad() {
    SomeClass.someMethod();
  }
}
```

SamCarlberg:
Fixes #410

### DIFF
--- a/app/app.gradle.kts
+++ b/app/app.gradle.kts
@@ -29,7 +29,6 @@ dependencies {
     compile(group = "de.huxhorn.lilith", name = "de.huxhorn.lilith.3rdparty.junique", version = "1.0.4")
     compile(group = "com.github.samcarlberg", name = "update-checker", version = "+")
     compile(group = "com.github.zafarkhaja", name = "java-semver", version = "0.9.0")
-    testCompile(project("test_plugins"))
 }
 
 val theMainClassName = "edu.wpi.first.shuffleboard.app.Shuffleboard"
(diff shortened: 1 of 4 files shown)


app/src/main/java/edu/wpi/first/shuffleboard/app/ShuffleboardUpdateChecker.java
```diff
@@ -218,8 +218,11 @@ public static void downloadNewestRelease(URL newestVersionLocation,
   private static File downloadFile(URL remoteFile, DoubleConsumer progressNotifier) throws IOException {
     File temp = File.createTempFile("newestshuffleboard", ".jar");
     log.finer("Downloading to " + temp);
+    URLConnection connection = remoteFile.openConnection();
+    connection.setReadTimeout(5000);    // ms
+    connection.setConnectTimeout(5000); // ms
     try (OutputStream out = new FileOutputStream(temp)) {
-      try (InputStream in = remoteFile.openStream()) {
+      try (InputStream in = connection.getInputStream()) {
```
AustinShalit: Not related to this PR but is there a reason this was not a try with multiple resources statement?

SamCarlberg: I find it less readable

AustinShalit: 👍

SamCarlberg:
Stale

PR #432, opened 2018-03-13
Make recording more efficient

### LINKED ISSUE #429: Long recordings use a lot of memory
The current implementation keeps track of all seen data points during the recording session, then serializes them all and saves to disk. This is inefficient because it re-serializes the same data points and rewrites the entire file from scratch every time the recording is saved, instead of only serializing the newest data since the most recent save event and updating the recording file.

Over time, hundreds of megabytes of data gets serialized and then immediately discarded, causing a _lot_ of GC pressure and, on lower-memory systems, can force swap space to be used.

SamCarlberg:
Recordings are saved incrementally, only modifying the existing recording file with new data, instead of reserializing all the recorded data points and rewriting the recording file from scratch.

This _significantly_ improves both the memory and CPU impact of data recording

Fixes #429

### DIFF
--- a/api/src/main/java/edu/wpi/first/shuffleboard/api/sources/recording/Recorder.java
+++ b/api/src/main/java/edu/wpi/first/shuffleboard/api/sources/recording/Recorder.java
@@ -16,6 +16,7 @@ import java.nio.file.Path;
 import java.time.Instant;
 import java.util.concurrent.Executors;
 import java.util.concurrent.TimeUnit;
+import java.util.logging.Level;
 import java.util.logging.Logger;
 
 import javafx.beans.property.BooleanProperty;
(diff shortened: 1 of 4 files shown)


api/src/main/java/edu/wpi/first/shuffleboard/api/sources/recording/Recorder.java
```diff
@@ -52,25 +62,33 @@ private Recorder() {
...
+      if (firstSave) {
+        Serialization.saveRecording(recording, file);
+        firstSave = false;
+      } else {
+        Serialization.updateRecordingSave(recording, file);
+      }
+      recording.getData().clear();
       log.fine("Saved recording to " + file);
-    } catch (IOException e) {
-      throw new RuntimeException("Could not save the recording", e);
```
SamCarlberg: This was removed because it would kill the auto-save worker

api/src/main/java/edu/wpi/first/shuffleboard/api/sources/recording/Serialization.java
```diff
@@ -83,35 +108,165 @@ public static void saveRecording(Recording recording, Path file) throws IOExcept
...
+      // Make sure it's a normal file (not a directory and not a symlink/shortcut to another file)
+      throw new IllegalArgumentException("Cannot insert into " + path);
+    }
+    Objects.requireNonNull(data, "Data cannot be null");
+    if (data.length == 0) {
+      // Nothing to insert, do nothing
+      return;
+    }
+    Path fileName = path.getFileName();
+    if (fileName == null) {
```
SamCarlberg: Have to do this check to appease FindBugs. This will never be null for a file, which we check for above

AustinShalit: What FindBugs warning is being thrown?

SamCarlberg: It was complaining about a possible null dereference. I don't remember the exact warning

PR #434, opened 2018-03-14
Add a check task to plugins to enforce up-to-date version numbers

SamCarlberg:
This will force future updates to plugins to bump the version number


SamCarlberg:
@JLLeitschuh Any thoughts on why travis wouldn't be able to run `git` commands? Might be a path thing?

JLLeitschuh:
Not sure, but in general, wouldn't this whole thing be easier with resource files?

SamCarlberg:
How would resource files be helpful? We still need to make sure the version numbers are bumped regardless of what files they're stored in

AustinShalit:
I agree with Jonathan.  He is saying that instead of parsing a java file, we should parse a resource file.  That way it is less likely to parse incorrectly.

JLLeitschuh:
Resource files can be modified/rewritten using the gradle task that already exists.
This is the example you want:
http://mrhaki.blogspot.com/2010/11/gradle-goodness-add-filtering-to.html

plugins/plugins.gradle.kts
```diff
@@ -3,6 +3,10 @@ import org.gradle.jvm.tasks.Jar
 
 subprojects {
     afterEvaluate {
+        /** The name of the plugin. This is, by convention, entirely lower-case. */
+        val pluginName: String = project.name
+        /** The plugin name as a proper noun, with the first character capitalized */
+        val proper = Character.toUpperCase(pluginName[0]) + pluginName.substring(1)
```
AustinShalit: Can we do something smarter here so it follows the name of the plugin rather than blindly capitalizing the first char?

SamCarlberg: That would be ideal. Project names are always lowercase eg `networktables`, `powerup`, `cameraserver` and cannot be reassigned. I suppose a property might be a good enough solution, even though it's not very enforceable

plugins/base/src/main/java/edu/wpi/first/shuffleboard/plugin/base/BasePluginVersion.java
```diff
@@ -0,0 +1,18 @@
...
+
+  /**
+   * The version of the Base plugin.
+   */
+  public static final String VERSION = "1.5.0";
+
+  private BasePluginVersion() {
+    throw new UnsupportedOperationException();
+  }
+}
```
AustinShalit: EOL @ EOF

plugins/cameraserver/src/main/java/edu/wpi/first/shuffleboard/plugin/cameraserver/CameraServerPluginVersion.java
```diff
@@ -0,0 +1,18 @@
...
+
+  /**
+   * The version of the CameraServer plugin.
+   */
+  public static final String VERSION = "1.5.0";
+
+  private CameraServerPluginVersion() {
+    throw new UnsupportedOperationException();
+  }
+}
```
AustinShalit: EOL @ EOF

plugins/networktables/src/main/java/edu/wpi/first/shuffleboard/plugin/networktables/NetworkTablesPluginVersion.java
```diff
@@ -0,0 +1,18 @@
...
+
+  /**
+   * The version of the NetworkTables plugin.
+   */
+  public static final String VERSION = "1.5.0";
+
+  private NetworkTablesPluginVersion() {
+    throw new UnsupportedOperationException();
+  }
+}
```
AustinShalit: EOL @ EOF

plugins/powerup/src/main/java/edu/wpi/first/shuffleboard/plugin/powerup/PowerUpPluginVersion.java
```diff
@@ -0,0 +1,18 @@
...
+
+  /**
+   * The version of the PowerUp plugin.
+   */
+  public static final String VERSION = "1.5.0";
+
+  private PowerUpPluginVersion() {
+    throw new UnsupportedOperationException();
+  }
+}
```
AustinShalit: EOL @ EOF

plugins/base/src/main/java/edu/wpi/first/shuffleboard/plugin/base/BasePluginVersion.java
SamCarlberg: Irrelevant for an autogenerated file IMO

AustinShalit: We should edit line six to say that it is autogenerated.

plugins/powerup/src/main/java/edu/wpi/first/shuffleboard/plugin/powerup/PowerUpPluginVersion.java
JLLeitschuh: You know spotless can enforce this right?

SamCarlberg:
Going to close this. It's too fragile. I'll add a PR template to prompt the author to check that they've updated the plugin versions

PR #438, opened 2018-03-17
Release opencv Mat after use to avoid memory leak

blucoat:
A new `Mat` gets created every time `ImageConverter.convert` gets called.  Now it gets released after the method is done.

On my laptop, this was causing shuffleboard to eat >5GB of memory after only a few minutes, when three camera streams were open at once.  I found that the Java heap was actually under a few hundred MB during this time.  My guess is that Java (at least, the OpenJDK implementation on my machine) didn't think it needed to garbage collect, since it wasn't running out of heap space for its own objects.  Thus it never collected the `Mat` objects that would get created each frame, so `finalize` never got called on them to call the C++ destructor.  In any case, this patch fixes it.

### DIFF
--- a/plugins/cameraserver/src/main/java/edu/wpi/first/shuffleboard/plugin/cameraserver/recording/serialization/ImageConverter.java
+++ b/plugins/cameraserver/src/main/java/edu/wpi/first/shuffleboard/plugin/cameraserver/recording/serialization/ImageConverter.java
@@ -23,7 +23,7 @@ public final class ImageConverter {
   private WritableImage image;
   private byte[] data;
   private ByteBuffer buffer;
-  private Mat converted = new Mat();
+  private final Mat converted = new Mat();
 
   /**
    * Convert a BGR-formatted OpenCV {@link Mat} into a JavaFX {@link Image}. JavaFX understands ARGB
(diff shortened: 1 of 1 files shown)


SamCarlberg:
This code was copy-pasted from GRIP, but shuffleboard has a different usecase. The `desiredHeight` parameter is used in GRIP to resize preview images in the pipeline to fit the preview pane; shuffleboard doesn't have that kind of restriction, since image size is determined on a widget-by-widget case.

The solution for shuffleboard is to remove the `desiredHeight`, and therefore the temporary `toRender` mat, altogether

blucoat:
Isn't it still needed to do the BGR to RGB conversion?

SamCarlberg:
No. The `toRender` mat is just a container for the resized image before copying to the JavaFX image. Since there's no resizing going on, it's redundant

blucoat:
So is the solution to subclass `PixelFormat` for BGR and single-channel images, so that `setPixels` can copy the image in one swoop?

SamCarlberg:
Pixel formats aren't designed to be extendable. There is a predefined enum for possible image types, each of which has its own read-only and writable formats. Barring reflection or bytecode hacks, the best we can do is do a gray->RGB conversion

blucoat:
Is this good to go?  Do you want me to squash this to one commit?

JLLeitschuh:
@SamCarlberg Should this change be ported to GRIP?

Issue #444, opened 2018-03-19
Slow startup on some machines


SamCarlberg:
Some users report very slow startup times, up to several minutes. This is Not Good.

For reference, it takes 4-5 seconds to load on my laptop (i7-6700HQ, mid-range SSD) and ~2.8 seconds on my desktop (i7-7700k, 960 Pro NVMe SSD).

In the past, I've seen the camera server plugin take a long time to load. I have not seen that behavior for a while, though

JLLeitschuh:
Asynchronous loading of some things perhaps?

SamCarlberg:
Do you mean that some things are being loaded asynchronously that shouldn't be, or that some things should be async that currently aren't?

JLLeitschuh:
If everything is loaded in the thread that calls `start` that might not be the best idea.

AustinShalit:
Closing as #466 has made this much better.

PR #445, opened 2018-03-19
Add a preloader to notify users of the status of startup routines

SamCarlberg:
This should help with #444 by letting users and CSAs know where Shuffleboard is during its startup sequence (it's already logged, but this is a better UX)

# Screenshot
image

### DIFF
--- a/app/src/main/java/edu/wpi/first/shuffleboard/app/Shuffleboard.java
+++ b/app/src/main/java/edu/wpi/first/shuffleboard/app/Shuffleboard.java
@@ -123,8 +123,6 @@ public class Shuffleboard extends Application {
     PluginLoader.getDefault().loadAllJarsFromDir(Storage.getPluginPath());
     notifyPreloader(new ShuffleboardPreloader.StateNotification("Loading custom plugins", 0.75));
     PluginCache.getDefault().loadCache(PluginLoader.getDefault());
-    notifyPreloader(new ShuffleboardPreloader.StateNotification("Loading custom plugins", 0.875));
-    Thread.sleep(250); // wait to let the new status be visible before overwriting it
     notifyPreloader(new ShuffleboardPreloader.StateNotification("Starting up", 1));
     Thread.sleep(20); // small wait to let the status be visible - the preloader doesn't get notifications for a bit
   }
--- a/app/src/main/java/edu/wpi/first/shuffleboard/app/Shuffleboard.java
+++ b/app/src/main/java/edu/wpi/first/shuffleboard/app/Shuffleboard.java
@@ -15,6 +15,7 @@ import edu.wpi.first.shuffleboard.plugin.powerup.PowerupPlugin;
 
 import com.github.zafarkhaja.semver.Version;
 import com.google.common.base.Stopwatch;
+import com.sun.javafx.application.LauncherImpl;
 
 import de.codecentric.centerdevice.javafxsvg.SvgImageLoaderFactory;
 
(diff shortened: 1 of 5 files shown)


app/src/main/java/edu/wpi/first/shuffleboard/app/Shuffleboard.java
```diff
@@ -110,25 +108,43 @@ public void start(Stage primaryStage) throws IOException {
...
+    notifyPreloader(new ShuffleboardPreloader.StateNotification("Loading CameraServer plugin", 0.375));
     PluginLoader.getDefault().load(new CameraServerPlugin());
+    notifyPreloader(new ShuffleboardPreloader.StateNotification("Loading Powerup plugin", 0.5));
     PluginLoader.getDefault().load(new PowerupPlugin());
+    notifyPreloader(new ShuffleboardPreloader.StateNotification("Loading custom plugins", 0.625));
     PluginLoader.getDefault().loadAllJarsFromDir(Storage.getPluginPath());
+    notifyPreloader(new ShuffleboardPreloader.StateNotification("Loading custom plugins", 0.75));
     PluginCache.getDefault().loadCache(PluginLoader.getDefault());
+    notifyPreloader(new ShuffleboardPreloader.StateNotification("Loading custom plugins", 0.875));
+    Thread.sleep(250); // wait to let the new status be visible before overwriting it
```
AustinShalit: Nah.... lets be faster every second counts

app/src/main/java/edu/wpi/first/shuffleboard/app/Shuffleboard.java
```diff
@@ -110,25 +108,43 @@ public void start(Stage primaryStage) throws IOException {
...
+    notifyPreloader(new ShuffleboardPreloader.StateNotification("Starting up", 1));
+    Thread.sleep(20); // small wait to let the status be visible - the preloader doesn't get notifications for a bit
```
AustinShalit: Every second counts.

app/src/main/java/edu/wpi/first/shuffleboard/app/Shuffleboard.java
```diff
@@ -57,8 +58,12 @@
 
   private final Stopwatch startupTimer = Stopwatch.createStarted();
 
+  public static void main(String[] args) {
+    LauncherImpl.launchApplication(Shuffleboard.class, ShuffleboardPreloader.class, args);
```
AustinShalit: How does this affect JUnique?  Should we need to acquire the lock here?

SamCarlberg: Doing the JUnique stuff in `init` lets us show a warning dialog in the future

app/src/main/java/edu/wpi/first/shuffleboard/app/Shuffleboard.java
SamCarlberg: Without a delay here, the final status will never be shown. It'll be stuck showing "Loading custom plugins" when that's already been completed

app/src/main/java/edu/wpi/first/shuffleboard/app/Shuffleboard.java
AustinShalit: Ok

app/src/main/java/edu/wpi/first/shuffleboard/app/Shuffleboard.java
AustinShalit: Ok

PR #452, opened 2018-03-22
Load main FXML in init()

SamCarlberg:
Avoids long downtimes when FXML loading/initialization takes a while, since that would happen in start() before the main stage was shown

### DIFF
--- a/app/src/main/java/edu/wpi/first/shuffleboard/app/MainWindowController.java
+++ b/app/src/main/java/edu/wpi/first/shuffleboard/app/MainWindowController.java
@@ -204,8 +204,6 @@ public class MainWindowController {
           .collect(joining(this::generateSeparatorLabel));
       connectionIndicatorArea.getChildren().setAll(collect);
     });
-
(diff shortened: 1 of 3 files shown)


JLLeitschuh:
I don't know enough about this to provide useful feedback. Does this improve the user experience?

AustinShalit:
>I don't know enough about this to provide useful feedback. Does this improve the user experience?

This will load the FXML files while the preloader is being shown rather than in the time between when the preloader is closed and the main window has yet to open.

app/src/main/java/edu/wpi/first/shuffleboard/app/MainWindowController.java
```diff
@@ -596,6 +594,9 @@ private void newTab() {
 
   @FXML
   private void showPlugins() {
+    if (pluginStage == null) {
+      setUpPluginsStage();
```
SamCarlberg: This was moved to here because the FXML init is now called on the launcher thread, not the application thread

PR #456, opened 2018-04-02
Fix plugin dependencies comparing versions in the wrong order

SamCarlberg:
Plugins could be unloaded without unloading their dependents first. Wasn't noticed before because plugins never changed their versions


JLLeitschuh:
Seems like something that would be valuable to have a unit test for.

PR #457, opened 2018-04-03
Fix occasional startup exceptions caused by running on launcher thread

SamCarlberg:
Specifically, source tree modifications should only be run on the application thread
Also forces source panes to be sorted by title


app/src/main/java/edu/wpi/first/shuffleboard/app/MainWindowController.java
```diff
@@ -327,6 +331,7 @@ private void setup(Plugin plugin) {
   private void tearDown(Plugin plugin) {
```
SamCarlberg: Note that `tearDown` does not need `FxUtils.runOnFxThread` since only `setup` can be called from the launcher thread in `Shuffleboard.init`

PR #458, opened 2018-04-03
Add a grid layout

SamCarlberg:
- Moves a lot of common layout code to `LayoutBase`
- Allows labels in List Layout to be moved to positions other than below their component
  - With labels on the left side, lists essentially become single-column grids
- Gives layouts the ability to define user-customizable properties like widgets already do
- Reduce the heights of various widgets to remove dead space
- Components can be dragged from layouts into the widget pane or into other layouts


api/src/main/java/edu/wpi/first/shuffleboard/api/plugin/Plugin.java
```diff
@@ -213,6 +214,13 @@ public void onUnload() {
     return ImmutableList.of();
   }
 
+  /**
+   * Gets the custom JSON serializers that this plugin requires to properly save and load custom components.
+   */
+  public List<ElementTypeAdapter<?>> getSavers() {
```
SamCarlberg: Naming is bad. @JLLeitschuh any recommendations?

JLLeitschuh: Converters?

api/src/main/java/edu/wpi/first/shuffleboard/api/util/ListUtils.java
```diff
@@ -21,6 +23,87 @@ private ListUtils() {
...
+    return Collector.<T, List<T>, ImmutableList<T>>of(
+        ArrayList::new,
+        List::add,
+        (left, right) -> {
+          left.addAll(right);
+          return left;
+        },
+        ImmutableList::copyOf
+    );
+  }
```
JLLeitschuh: Guava 21.0+ has this already. `ImmutableList.toImmutableList()`.
http://google.github.io/guava/releases/23.0/api/docs/com/google/common/collect/ImmutableList.html#toImmutableList--

PR #466, opened 2018-05-08
Improve startup times and make the preloader nicer 

SamCarlberg:
- Lazily initialize FXML helper dialogs (plugins pane, various update prrompts, "about" dialog) instead of loading them all at startup
  - Reduction of ~30% in FXML load times
- Do not start a NetworkTable client before setting the server ID (which itself starts the clients)

Additionally, using pre-extracted JNI libraries reduces NetworkTable load times from ~260ms to ~10ms and CameraServer from ~450ms to ~10ms on my desktop computer, and total startup time from ~3.3 seconds to ~2.7 seconds. This PR does not add JNI pre-extraction, since ntcore and cscore will handle that process.

Helps with #444, but that will probably be most affected by using pre-extracted JNI

# Screenshot

screenshot from 2018-04-27 16-24-52

image


AustinShalit:
Can we add the WPILib logo to the preloader while we are doing this?

api/src/test/java/edu/wpi/first/shuffleboard/api/util/LazyInitTest.java
```diff
@@ -0,0 +1,39 @@
...
+    LazyInit<Object> throwing = LazyInit.of(() -> {
+      throw new UnsupportedOperationException();
+    });
+    try {
+      throwing.get();
+      fail("No exception was thrown");
+    } catch (RuntimeException e) {
+      assertNotNull(e.getCause());
+      assertEquals(UnsupportedOperationException.class, e.getCause().getClass());
+    }
```
AustinShalit: Is it possible to use an `assertThrows` here?

api/src/test/java/edu/wpi/first/shuffleboard/api/util/LazyInitTest.java
```diff
@@ -0,0 +1,39 @@
...
+  @Test
+  public void testInitializerThrows() {
+    LazyInit<Object> throwing = LazyInit.of(() -> {
+      throw new UnsupportedOperationException();
+    });
+    try {
+      throwing.get();
+      fail("No exception was thrown");
+    } catch (RuntimeException e) {
+      assertNotNull(e.getCause());
```
AustinShalit: I think this assert is redundant because the equals below will report e as null.

app/src/main/java/edu/wpi/first/shuffleboard/app/HexagonGrid.java
```diff
@@ -0,0 +1,63 @@
...
+
+import java.util.Collection;
+import java.util.stream.Collectors;
+
+import javafx.scene.Group;
+import javafx.scene.shape.Polygon;
+
+public class HexagonGrid extends Group {
+
+  private static final double ang30 = Math.toRadians(30);
```
AustinShalit: Kinda surprised checkstyle did not catch this... should be named `ANGLE_30`.

app/src/main/java/edu/wpi/first/shuffleboard/app/PreloaderController.java
```diff
@@ -4,16 +4,26 @@
...
 import javafx.fxml.FXML;
 import javafx.scene.control.Label;
 import javafx.scene.control.ProgressBar;
 import javafx.scene.layout.Pane;
+import javafx.scene.paint.Color;
 
 public class PreloaderController {
 
+  private static final Color primaryColor = Color.rgb(39, 39, 247);
+  private static final Color secondaryColor = Color.rgb(64, 64, 255);
```
AustinShalit: How do you feel about using colors that complement our branding guide?

https://github.com/wpilibsuite/branding

app/src/main/java/edu/wpi/first/shuffleboard/app/PreloaderController.java
```diff
@@ -22,13 +32,31 @@
...
   @FXML
   private void initialize() {
+    // Bring the hexagons to the top edge to avoid weird blank spots
+    backgroundContainer.setTranslateY(-HEXAGON_RADIUS);
+
     progressBar.setProgress(-1);
     versionLabel.setText(Shuffleboard.getVersion());
+
+    // Hexagon grid background
+    Random random = new Random(System.currentTimeMillis() ^ (System.currentTimeMillis() >> 16));
```
AustinShalit: Just curious why you are using a custom seed...

api/src/test/java/edu/wpi/first/shuffleboard/api/util/LazyInitTest.java
SamCarlberg: No, because the underlying exception is wrapped. Hence the check against the cause instead of the exception itself

AustinShalit: assertThrows returns the exception, so you should be able to grab it.  tldr: I think `fail` is ugly here and am trying to get rid of it.

```Java
@Test
void exceptionTesting() {
    Throwable exception = assertThrows(IllegalArgumentException.class, () -> {
        throw new IllegalArgumentException("a message");
    });
    assertEquals("a message", exception.getMessage());
}
```

app/src/main/java/edu/wpi/first/shuffleboard/app/PreloaderController.java
SamCarlberg: This color scheme matches the default theme.  It'd be a bit weird to have a red-and-gray preloader switch to a blue-and-white theme when shuffleboard is launched.

If the WPILib logo is added, I would switch to a red color scheme.  I don't know where the logo is actually used besides the developer shirts; it's definitely not in any other desktop tool, so it's not something that people would recognize and immediately recognize as "WPILib"

app/src/main/java/edu/wpi/first/shuffleboard/app/PreloaderController.java
SamCarlberg: Because I took this straight from GRIP and didn't change it

app/src/main/java/edu/wpi/first/shuffleboard/app/PreloaderController.java
ThadHouse: The installer will have it, and vscode as the extension icon, but it won't be anywhere actually in the extension

PR #486, opened 2018-06-22
Update issue templates

SamCarlberg:
Add templates for bug reports, feature requests

Modified spotless to allow two trailing spaces on lines in markdown files, since that's what markdown uses to denote line breaks

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -1,4 +1,5 @@
 
+import com.diffplug.spotless.FormatterStep
 import edu.wpi.first.wpilib.versioning.ReleaseType
 import org.gradle.api.Project
 import org.gradle.api.plugins.quality.FindBugs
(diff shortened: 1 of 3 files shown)


build.gradle.kts
```diff
@@ -46,7 +47,46 @@ allprojects {
...
+                        // Trim the excess, but leave two spaces at the end for a tight line break
+                        return match.groupValues[1] + "  "
+                    }
+                    if (line.endsWith(" ")) {
+                        // Ends with a single space - remove it
+                        return line.substring(0, line.length - 1)
+                    }
+                    // Not sure how we got here; every case should have been covered.
+                    // Print an error but do not change the line
+                    System.err.println("Could not trim whitespace from line '$line'")
```
JLLeitschuh: You can throw an exception which would fail the build. Up to you though.

PR #488, opened 2018-06-28
Fix multi-byte characters in strings in recordings

### LINKED ISSUE #487: Loading recordings barfs on strings containing code points > 255
**Describe the bug**  
Data recordings store characters in strings as 1-byte code points. Multi-byte characters (eg ™, ®, non-Latin letters) are saved correctly, but treated as 1-byte when calculating the length of the string

**To Reproduce**  
1. Have a source containing a string with a code point > 255, such as ®
2. Attempt to load the recording

SamCarlberg:
All chars were assumed to be 1 byte, causing issues with characters outside ASCII

Fixes #487

### DIFF
--- a/api/src/test/java/edu/wpi/first/shuffleboard/api/sources/recording/SerializationTest.java
+++ b/api/src/test/java/edu/wpi/first/shuffleboard/api/sources/recording/SerializationTest.java
@@ -13,7 +13,6 @@ import java.util.List;
 import static org.junit.jupiter.api.Assertions.assertArrayEquals;
 import static org.junit.jupiter.api.Assertions.assertEquals;
 
-@SuppressWarnings("PMD")
 public class SerializationTest {
 
   private static final byte[] fooBarBytes = new byte[]{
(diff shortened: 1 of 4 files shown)


api/src/test/java/edu/wpi/first/shuffleboard/api/sources/recording/SerializationTest.java
```diff
@@ -22,6 +22,10 @@
       0, 0, 0, 3, 'b', 'a', 'r'  // "bar", encoded with length
   };
 
+  // Grinning emoji, four bytes
+  @SuppressWarnings("AvoidEscapedUnicodeCharacters")
+  private static final String grinningEmoji = "\uD83D\uDE01";
```
AustinShalit: Doesn't java support emojis in code now?  Can we put the emoji here instead of the escape sequence?

SamCarlberg: Huh, I didn't know that. That'll make it easier

AustinShalit:

AustinShalit:
@frcjenkins test this please

SamCarlberg:

PR #489, opened 2018-06-29
Fix occasional graph widget NPEs

SamCarlberg:
Was caused by adding newly created graph widgets to the pool before they were initialized, resulting in null pointers on FXML-injected fields


JLLeitschuh:
It was problems like this that pushed us towards using a DI framework for GRIP.
Using DI allowed us to only expose instances of objects after they were fully constructed and the FXML loader had fully injected everything.

PR #492, opened 2018-07-18
Allow tabs to be defined and populated remotely

### LINKED ISSUE #473: Layouts should be able to be defined from robot code
See wpilibsuite/allwpilib#1022

Tabs defined by robots should be:

- Uneditable (cannot add, move, resize, or remove tiles)
- Undeletable
- Color-coded to be visually distinct from user-defined tabs

Robot-defined tabs should be generated when the NetworkTables plugin connects to a client. These tabs should be deleted when connecting to a new client - but _not_ when disconnected, since that could happen if a robot or radio reboots mid-match, and tabs disappearing on the drivers would be terrible UX.

SamCarlberg:
Relies on wpilibsuite/allwpilib#1022

Adds a new tab type for tabs whose contents and layouts are defined remotely instead of by the user. These tabs have special coloring - by default, a darker background and yellow text - and will _not_ be saved as part of the overall layout. Default tabs defined by `Plugin.getDefaultTabInfo()` and user-created tabs though the UI are not changed and will still be saved as usual.

# Screenshots

## Initial view
Note the "Drivers" and "Commands" tabs have different styles versus the normal tabs
screenshot from 2018-07-18 10 32 18

## Simple widget with position and size defined through NetworkTables
screenshot from 2018-07-18 10 32 25

## Complex widgets and layouts
screenshot from 2018-07-18 10 32 36

Closes #473


SamCarlberg:
I'm going to merge this as-is. The spec in WPILib is dialed in and is waiting on C++ support, but having the capability present in Shuffleboard will allow third-party plugins to start getting familiar with the new API

PR #493, opened 2018-07-23
Fix regression where newly added tiles are not resizable

SamCarlberg:
Tiles would only be resizable after they're dragged around


AustinShalit:
Is there a test for this to make sure it does not happen again?

SamCarlberg:
This is difficult to write tests for; testfx cannot do drags when headless

PR #494, opened 2018-07-23
Fix regression with context menus on interactive source trees

SamCarlberg:
When changing the dashboard (i.e. creating a new layout or loading a saved layout), callbacks would still reference the original dashboard instead of the newly loaded one


AustinShalit:
Is there a test for this to make sure it does not happen again?

SamCarlberg:
It would be difficult, since testfx cannot do context menus when headless

PR #496, opened 2018-08-01
Save window position and size

SamCarlberg:
Loading a save file will set the size and position of the window. If an older save file is loaded without window data, the file will be loaded normally but the window's position and size will not change. If the saved position would place the window off-screen, it will be resized but not moved.

### DIFF
--- a/app/src/main/java/edu/wpi/first/shuffleboard/app/DashboardData.java
+++ b/app/src/main/java/edu/wpi/first/shuffleboard/app/DashboardData.java
@@ -9,19 +9,19 @@ public final class DashboardData {
 
   private final double dividerPosition;
   private final DashboardTabPane tabPane;
-  private final WindowData windowData;
+  private final WindowGeometry windowGeometry;
 
   /**
    * Creates a new dashboard data object.
    *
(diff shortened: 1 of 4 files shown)


app/src/main/java/edu/wpi/first/shuffleboard/app/DashboardData.java
```diff
@@ -7,34 +7,32 @@
...
 
   /**
    * Creates a new dashboard data object.
    *
    * @param dividerPosition the position of the divider between the dashboard tabs and the sources/gallery pane
    * @param tabPane         the dashboard tab pane
+   * @param windowData      data for the window
    */
-  public DashboardData(double dividerPosition, DashboardTabPane tabPane) {
+  public DashboardData(double dividerPosition, DashboardTabPane tabPane, WindowData windowData) {
```
AustinShalit: How about `WindowGeometry` instead?  Data is very abstract...

SamCarlberg:
@AustinShalit I've been testing this on my 3-monitor desktop. Window size and location are set appropriately.

AustinShalit:
ok.  go ahead

PR #497, opened 2018-08-03
Update controllable sendables

SamCarlberg:
Speed controller, relay, and drive base widgets will be disabled if the `.controllable` entry is set to `false`

Remove the LiveWindow helper class, since that only existed to disable actuator widgets when not in test mode - that information is now available to every widget.


SamCarlberg:

PR #508, opened 2018-08-22
Make builds for single platforms

SamCarlberg:
Reduces JAR size from 105MB to about 50MB (per platform)  
Remove cedarsoft semver library, since it was pulling in the kotlin runtime (and some other libraries).  
Refactor PluginLoader a bit to avoid a new God Class classification

By default, the build will use native artifacts for the running operating system. This can be changed with the `platform` gradle property:

| Target OS | Flag |
|---|---|
| Windows 32-bit | `-Pplatform=win32` |
| Windows 64-bit | `-Pplatform=win64` |
| Mac OS (10.7+)  | `-Pplatform=mac64` |
| Liinux 64-bit | `-Pplatform=linux64` |
 
Platform builds for the app fat jar use the platform names above for the artifact classifier.

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -100,28 +100,6 @@ allprojects {
     }
 }
 
-val currentPlatform: String
-    get() {
-        val os: String = when {
-            System.getProperty("os.name").contains("windows", true) -> "win"
-            System.getProperty("os.name").contains("mac", true) -> "mac"
-            System.getProperty("os.name").contains("linux", true) -> "linux"
-            else -> throw UnsupportedOperationException("Unknown OS: ${System.getProperty("os.name")}")
-        }
-        val osarch = System.getProperty("os.arch")
-        var arch: String =
-                if (osarch.contains("x86_64") || osarch.contains("amd64")) {
-                    "64"
-                } else if (osarch.contains("x86")) {
-                    "32"
-                } else {
-                    throw UnsupportedOperationException(osarch)
-                }
-        return os + arch
-    }
-
-println("Current OS: $currentPlatform")
-
 allprojects {
     apply {
         plugin("java")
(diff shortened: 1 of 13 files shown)


api/src/test/java/edu/wpi/first/shuffleboard/api/sources/recording/SerializationTest.java
```diff
@@ -158,7 +158,7 @@ private void deleteFile(Path file) throws IOException {
     if (System.getenv("CI") != null) {
       return;
     }
-    Files.delete(file);
+    //Files.delete(file);
```
AustinShalit: Is this right?

build.gradle.kts
```diff
@@ -114,6 +115,10 @@ allprojects {
         mavenCentral()
     }
 
+    project.ext {
+        this["platform"] = properties["platform"] ?: currentPlatform
+    }
```
JLLeitschuh: This could be:
```kotlin
val platform : String by project.extra {
    project.properties["platform"].toString() ?: currentPlatform
}
```
(You may be able to just do `by extra` without the `project`, idk, try it).

That would allow your code below to be simplified because `platform` would be in scope:
```kotlin
tasks.withType<ShadowJar> {
    classifier = platform
}
```

SamCarlberg: This sets a property on the project so the subprojects know what platform to get artifacts for.
It also seems to be the only way to get the script to compile without an exception:
```
FAILURE: Build failed with an exception.
* What went wrong:
A problem occurred configuring root project 'shuffleboard'.
> Cannot configure the 'publishing' extension after it has been accessed.
```

JLLeitschuh: Ah, interesting, then `val platform  : String by project.ext` should work exactly the same.
If it doesn't something's wrong.

SamCarlberg: Work where? That's a property use, not a property assignment. It also wouldn't be in scope for the shadowjar task anyway, since that's in a separate project configuration block.

JLLeitschuh: Oh, I opened it up and it looked like it was in the same block. My bad.

SamCarlberg:

Issue #400, opened 2018-02-05
Make widget title bar optional


andrewda:
It would be nice to have the option to disable title bars so that they don't take up unnecessary space. If we've got a lot of information to display, every inch is valuable, so the smaller we can make some widgets the better.

AustinShalit:
Duplicate of #42

andrewda:
Should have made this more clear in my original issue, but I'm talking about the title bars on widgets here. It would be nice to be able to disable the widget titles on some specific widgets if I want, for example, a long boolean box across the top of the window. Having the option to disable the title bar of that boolean box would reduce the footprint of the widget on the screen and allow for a cleaner UI.

I _believe_ #42 is specifically referring to hiding the larger UI features and titles, not the titles of specific widgets.

PR #520, opened 2018-10-01
Make widget title bars optional

andrewda:
Closes https://github.com/wpilibsuite/shuffleboard/issues/400

# Overview
This PR adds a widget property allowing the title bar to be hidden. This is useful for widgets which don't need a title or are fairly self-explanatory, and to save room on the screen.

# Screenshots
With widget title bar:
with_title

Without widget title bar:
without_title

### DIFF
--- a/api/src/main/java/edu/wpi/first/shuffleboard/api/widget/AbstractWidget.java
+++ b/api/src/main/java/edu/wpi/first/shuffleboard/api/widget/AbstractWidget.java
@@ -9,6 +9,8 @@ import com.google.common.collect.ImmutableList;
 import java.util.List;
 
 import javafx.beans.InvalidationListener;
+import javafx.beans.property.BooleanProperty;
+import javafx.beans.property.SimpleBooleanProperty;
 import javafx.beans.property.SimpleStringProperty;
 import javafx.beans.property.StringProperty;
 import javafx.beans.value.ChangeListener;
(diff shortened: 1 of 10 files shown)


andrewda:
I definitely understand where you're coming from there, but I feel like teams should be able to organize their dashboards how they see fit. Also, in my experience, it's rare to change our dashboard layout after initially setting it up. If we needed to move one of those widgets that's difficult to move without a title bar, we can very easily toggle it back on and drag it around.

I also find it weird to limit the components that can have their titles disabled. If a team finds a use for disabling certain titles, why should they be limited from doing so? If it doesn't make sense for a title to be disabled, then people won't disable it, but it leaves the option open for a cleaner experience in certain scenarios.

My team has used this for a few events, and it has definitely been a nice addition to have, especially when cramming a bunch of widgets on a small screen. Drivers get used to the dashboard's layout, and don't need to read the title of every widget mid-competition. Obviously, switches and buttons will almost always have a title/label, but having a big red/green boolean box across the top of the screen is incredibly useful to display important robot information. For example, which of the following is more visible to a driver glancing down for a fraction of a second, looking for a known color:

|With title|Without title|
|--------|-------|
|image|image|

Thanks for taking the time to try it out, though! Hopefully this makes it in sometime soon, and in the meantime we'll just continue using this branch. Let me know if there's anything else I can do/change!

SamCarlberg:
You make some good points, but we'll have to take some time to figure out if this is something good for the future of shuffleboard. If we decide to merge, it'll be after the 2019 season is over. We're trying to only merge bug fixes or minor improvements during the season this year.

virtuald:
This seems really useful.

PeterJohnson:
Superseded by #606.

PR #523, opened 2018-10-04
Build releases for all platforms with a single task

SamCarlberg:
# Overview

Exposes a task on the `app` project to build native JARs for each supported native platform.

Calling `./gradlew :app:shadowJarAllPlatforms` will generate the platform-specific JAR files in `app/build/libs`

Maven publishing will publish all the native app JARs, regardless of the build platform

Removes the `-Pplatform=...` flag. Use the platform-specific tasks to build the release JARs for those platforms, or use `./gradlew :app:shadowjarAllPlatforms` to build all of them at once

---

Update checkstyle to it won't freaking crash every time it tries to find the suppressions file


buildSrc/src/main/kotlin/NativeConfigurations.kt
```diff
@@ -0,0 +1,89 @@
...
+ *
+ * @param path the path to the project, e.g. `":plugins:networktables"`
+ */
+fun DependencyHandlerScope.nativeProject(path: String) {
+    forEachPlatform {
+        nativeProject(path, it)
+    }
+}
+
+internal fun DependencyHandlerScope.nativeProject(path: String, platform: NativePlatforms) {
```
JLLeitschuh: You can probably make this `DependencyHandler` instead of `DependencyHandlerScope`, I might be wrong though.

checkstyle.xml
```diff
@@ -23,7 +23,7 @@
     <property name="fileExtensions" value="java, properties, xml"/>
 
     <module name="SuppressionFilter">
-        <property name="file" value="checkstyleSuppressions.xml"/>
+        <property name="file" value="${config_loc}/checkstyleSuppressions.xml"/>
```
JLLeitschuh: 👍

SamCarlberg: This breaks Windows because of course it does

app/app.gradle.kts
```diff
@@ -65,6 +65,70 @@ tasks.withType<Test> {
...
+    }
+}
+
+tasks.create("shadowJarAllPlatforms") {
+    nativeShadowTasks.forEach {
+        this.dependsOn(it)
+    }
+}
+
+tasks.withType<ShadowJar> {
```
JLLeitschuh: You really want to start using the deferred configuration API.
```kotlin
tasks.withType<ShadowJar>().configureEach {
    ...
```

app/app.gradle.kts
```diff
@@ -65,6 +65,70 @@ tasks.withType<Test> {
...
+                project.configurations.compile,
+                project.configurations.getByName(platform.platformName)
+        )
+        from(
+                java.sourceSets["main"].output,
+                project(":api").java.sourceSets["main"].output,
+                project(":plugins:base").java.sourceSets["main"].output,
+                project(":plugins:cameraserver").java.sourceSets["main"].output,
+                project(":plugins:networktables").java.sourceSets["main"].output,
+                project(":plugins:powerup").java.sourceSets["main"].output
```
JLLeitschuh: You could probably make this more generic and less copy pasted.

app/app.gradle.kts
SamCarlberg: What's the advantage?

app/app.gradle.kts
```diff
@@ -65,6 +65,70 @@ tasks.withType<Test> {
     dependsOn(project("test_plugins").tasks["jar"])
 }
 
+val nativeShadowTasks = NativePlatforms.values().map { platform ->
+    tasks.create<ShadowJar>("shadowJar-${platform.platformName}") {
```
JLLeitschuh: Consider using `tasks.register<ShadowJar>("shadowJar-${platform.platformName}")` instead.

app/app.gradle.kts
ThadHouse: configureEach uses the lazy task configuration API, which enables quicker builds, and at some point in the future will become the only way to configure tasks.

SamCarlberg: Oh nice :+1:

app/app.gradle.kts
SamCarlberg: I wish.  I was trying to get it into the `nativeProject` function but the DependencyHandler class doesn't have access to the enclosing project

app/app.gradle.kts
SamCarlberg: `register` only lets you configure a default Task, not a ShadowJar task like I need here

SamCarlberg: It's introduced in 4.10, which has some API compatibility breaks with 4.9

JLLeitschuh: You can still use `tasks.register("shadowJar-${platform.platformName}", Shadow::class.java)` in 4.9.

app/app.gradle.kts
JLLeitschuh:
```kotlin
val projectOutputs = 
     setOf(project(":api"), project(":plugins:base"), ....)
         .map { it.java.sourceSets["main"].output  } 
val outputs =
    setOf(java.sourceSets["main"].output) + projectOutputs
from(outputs)
```

build.gradle.kts
```diff
@@ -207,7 +208,14 @@ allprojects {
     }
 
     afterEvaluate {
+        // Note: plugins should override this
         version = getWPILibVersion() ?: getVersionFromGitTag()
+        tasks.withType<Jar> {
```
JLLeitschuh: Why in an `afterEvaluate`?

ThadHouse:
@SamCarlberg @JLLeitschuh I PR'd some version fixes into this repo, at https://github.com/SamCarlberg/shuffleboard/pull/1 Can you guys take a look at them and make sure I'm not horribly messing things up. Not familiar with the kts setup, but I tried my best.

app/app.gradle.kts
SamCarlberg: This doesn't seem to work with the publishing extension

plugins/plugins.gradle.kts
```diff
@@ -30,17 +23,11 @@ subprojects {
             create<MavenPublication>("plugin.${project.name}") {
                 groupId = "edu.wpi.first.shuffleboard.plugin"
                 artifactId = project.name
-                getWPILibVersion()?.let { version = it }
+                version = project.version as String
```
JLLeitschuh: Use `.toString()`.

SamCarlberg: I don't think that's necessary since all project versions are going to be strings

app/app.gradle.kts
JLLeitschuh: Oh, right, it will in Gradle 4.10.

I have a stopgap measure that works for gradle 4.9 but it uses an internal API.
```
/**
 * Stopgap measure to support [TaskProvider] when a [PublishArtifact] is needed.
 * [Issue](https://github.com/gradle/gradle/issues/6246).
 */
fun lazyPublishArtifactHack(provider: TaskProvider<*>): PublishArtifact =
    LazyPublishArtifact(provider)
```

Using it lets you do this:
```kotlin
            nativeShadowTasks.forEach {
                artifact(lazyPublishArtifactHack(it)) {
                    classifier = it.classifier
                }
            }
```

Otherwise, if you don't want to use the hack you can do:
```kotlin
            nativeShadowTasks.forEach {
                // TODO: Fix this for Gradle 5.0
                artifact(it.get()) {
                    classifier = it.classifier
                }
            }
```

SamCarlberg: I'd prefer to fix it when the API becomes available

JLLeitschuh: Okay.

azure-pipelines.yml
```diff
@@ -120,7 +120,7 @@ jobs:
         jdkDirectory: '/Library/Java/JavaVirtualMachines/jdk-11.jdk/Contents/Home/'
         publishJUnitResults: false
         testResultsFiles: '**/TEST-*.xml'
-        tasks: 'shadowJar test'
+        tasks: 'check'
```
JLLeitschuh: You probably just want to straight up run the `build` task which will just build everything and run the `check` task for you.

SamCarlberg: I think we're starting to stray too far from what this PR is for. Ci improvements can be made in separate PRs; the only reason the azure setup was modified here was because checkstyle got fixed in this PR

PR #538, opened 2018-11-05
Add markers to data recordings

SamCarlberg:
# Overview

This adds a mechanism for adding event markers to data recordings. Markers can be added to a recording from plugins with `Recorder.addMarker(name, description, importance)`.

## Creating markers via NetworkTables

Markers can also be added by NetworkTables clients (eg robot programs) by creating or updating an entry with the format
`/Shuffleboard/.recording/events/<marker name>/Info=["description", "importance"]`
The description is optional and may be an empty string.  
The importance must be one of `"TRIVIAL"`, `"LOW"`, `"NORMAL"`, `"HIGH"`, or `"CRITICAL"` (capitalization is ignored).

Recording converters and UI elements will be added in followup pull requests.

### DIFF
--- a/api/src/main/java/edu/wpi/first/shuffleboard/api/sources/recording/Marker.java
+++ b/api/src/main/java/edu/wpi/first/shuffleboard/api/sources/recording/Marker.java
@@ -1,0 +1,94 @@
+package edu.wpi.first.shuffleboard.api.sources.recording;
+
+import java.util.Objects;
+
+/**
+ * Marks an event in a recording.
+ */
+public final class Marker {
+
+  private final String name;
+  private final String description;
+  private final MarkerImportance importance;
+  private final long timestamp;
+
+  /**
+   * Creates a new event marker.
+   *
+   * @param name        the name of the marked event
+   * @param description a description of the marked event
+   * @param importance  the importance of the event
+   * @param timestamp   the time the event occurred, measured in milliseconds since the start of the recording
+   */
+  public Marker(String name, String description, MarkerImportance importance, long timestamp) {
+    this.name = Objects.requireNonNull(name, "name");
+    this.description = Objects.requireNonNull(description, "description");
+    this.importance = Objects.requireNonNull(importance, "importance");
+    this.timestamp = timestamp;
+  }
+
+  /**
+   * Creates a new event marker with no description.
+   *
+   * @param name       the name of the marked event
+   * @param importance the importance of the event
+   * @param timestamp  the time the event occurred, measured in milliseconds since the start of the recording
+   */
+  public Marker(String name, MarkerImportance importance, long timestamp) {
+    this(name, "", importance, timestamp);
+  }
+
+  /**
+   * Gets the name of the marked event.
+   */
+  public String getName() {
+    return name;
+  }
+
+  /**
+   * Gets the description of the marked event.
+   */
+  public String getDescription() {
+    return description;
+  }
+
+  /**
+   * Gets the importance of the marked event.
+   */
+  public MarkerImportance getImportance() {
+    return importance;
+  }
+
+  /**
+   * Gets the timestamp of the marked event.
+   */
+  public long getTimestamp() {
+    return timestamp;
+  }
+
+  @Override
+  public String toString() {
+    return String.format("Marker(name=\"%s\", description=\"%s\", importance=%s, timestamp=%d)",
+        name, description, importance, timestamp);
+  }
+
+  @Override
+  public boolean equals(Object obj) {
+    if (this == obj) {
+      return true;
+    }
+    if (obj == null || getClass() != obj.getClass()) {
+      return false;
+    }
+    Marker that = (Marker) obj;
+    return this.name.equals(that.name)
+        && this.description.equals(that.description)
+        && this.importance.equals(that.importance)
+        && this.timestamp == that.timestamp;
+  }
+
+  @Override
+  public int hashCode() {
+    return Objects.hash(name, description, importance, timestamp);
+  }
+}
--- a/api/src/main/java/edu/wpi/first/shuffleboard/api/sources/recording/MarkerImportance.java
+++ b/api/src/main/java/edu/wpi/first/shuffleboard/api/sources/recording/MarkerImportance.java
@@ -4,11 +4,16 @@ package edu.wpi.first.shuffleboard.api.sources.recording;
  * The levels of importance of a marker.
  */
 public enum MarkerImportance {
+
+
+  // Maintainer note: this enum is mirrored in WPILib. Adding or removing entries, or changing the name of any entry,
+  // requires a corresponding change to WPILib.
+
   /**
    * The lowest importance level. This can be used for marking the most minor events that are nice to know if something
    * goes wrong, but otherwise has no use.
    */
-  NONE,
+  TRIVIAL,
   /**
    * Marks an event with low importance.
    */
@@ -22,7 +27,7 @@ public enum MarkerImportance {
    */
   HIGH,
   /**
-   * Marks a critical event such as a component failure or robot reboot.
+   * Marks a critically important event such as a component failure, power loss, software deadlock, or timeout.
    */
   CRITICAL;
 
@@ -36,7 +41,7 @@ public enum MarkerImportance {
   public static MarkerImportance valueOf(int ordinal) {
     switch (ordinal) {
       case 0:
-        return NONE;
+        return TRIVIAL;
       case 1:
         return LOW;
       case 2:
--- a/api/src/main/java/edu/wpi/first/shuffleboard/api/sources/recording/MarkerImportance.java
+++ b/api/src/main/java/edu/wpi/first/shuffleboard/api/sources/recording/MarkerImportance.java
@@ -4,11 +4,27 @@ package edu.wpi.first.shuffleboard.api.sources.recording;
  * The levels of importance of a marker.
  */
 public enum MarkerImportance {
-  LOWEST,
+  /**
+   * The lowest importance level. This can be used for marking the most minor events that are nice to know if something
+   * goes wrong, but otherwise has no use.
+   */
+  NONE,
+  /**
+   * Marks an event with low importance.
+   */
   LOW,
+  /**
+   * Marks an event with no special connotations, such as the start of a match or match period, or scoring a game piece.
+   */
   NORMAL,
+  /**
+   * Marks an important event.
+   */
   HIGH,
-  HIGHEST;
+  /**
+   * Marks a critical event such as a component failure or robot reboot.
+   */
+  CRITICAL;
 
   /**
    * Identical to {@code values()[ordinal]}, but without the array copy.
@@ -20,7 +36,7 @@ public enum MarkerImportance {
   public static MarkerImportance valueOf(int ordinal) {
     switch (ordinal) {
       case 0:
-        return LOWEST;
+        return NONE;
       case 1:
         return LOW;
       case 2:
@@ -28,7 +44,7 @@ public enum MarkerImportance {
       case 3:
         return HIGH;
       case 4:
-        return HIGHEST;
+        return CRITICAL;
       default:
         throw new IllegalArgumentException("Ordinal out of range: " + ordinal);
     }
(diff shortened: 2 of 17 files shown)


api/src/main/java/edu/wpi/first/shuffleboard/api/sources/recording/MarkerImportance.java
```diff
@@ -0,0 +1,58 @@
+package edu.wpi.first.shuffleboard.api.sources.recording;
+
+/**
+ * The levels of importance of a marker.
+ */
+public enum MarkerImportance {
+
+
+  // Maintainer note: this enum is mirrored in WPILib. Adding or removing entries, or changing the name of any entry,
+  // requires a corresponding change to WPILib.
```
JLLeitschuh: Move to a common library?

SamCarlberg: Not really feasible since it has to be done in both C++ and Java WPILib, and I don't think the enum will change much (or at all), so the overhead of maintaining another (_tiny_) library just for this is not worth it IMO

api/src/test/java/edu/wpi/first/shuffleboard/api/sources/recording/SerializationTest.java
```diff
@@ -131,6 +132,20 @@ public void testUpdateFileWithNewDataTypes() throws IOException {
     assertEquals(data, last.getData(), "Second update was wrong");
   }
 
+  @Test
+  public void testEncodeRecodeWithMarkers() throws IOException {
+    final Path file = Files.createTempFile("testEncodeRecodeWithMarkers", ".sbr");
```
JLLeitschuh: Do you ever clean these files up? Are you certain you aren't leaking temporary files all over your file system doing this?

You should really use Junit Pioneer's temporary directory extension:
https://junit-pioneer.org/docs/temp-directory/

plugins/networktables/src/test/java/edu/wpi/first/shuffleboard/plugin/networktables/MarkerGeneratorTest.java
```diff
@@ -0,0 +1,109 @@
...
+
+  private void assertNoMarkersAdded() {
+    waitForEntry();
+    assertEquals(0, recorder.getRecording().getMarkers().size(), "No markers should have been added");
+  }
+
+  private void waitForEntry() {
+    recorder.start();
+    generator.start();
+    sleep(50, TimeUnit.MILLISECONDS);
```
JLLeitschuh: Wouldn't it be better to use something like this awaiter library instead of arbitrarily sleeping for some length of time?
https://github.com/jhalterman/concurrentunit

api/src/main/java/edu/wpi/first/shuffleboard/api/sources/recording/MarkerImportance.java
```diff
@@ -0,0 +1,67 @@
...
+  }
+
+  /**
+   * Gets the importance level with the given ID number.
+   *
+   * @param id the ID number of the enum constant to get
+   *
+   * @return the enum constant with the given ID
+   */
+  public static MarkerImportance forId(int id) {
```
JLLeitschuh: Much better!

api/src/main/java/edu/wpi/first/shuffleboard/api/sources/recording/Recorder.java
```diff
@@ -221,4 +237,15 @@ public StringProperty fileNameFormatProperty() {
...
+  /**
+   * Gets the recording being recorded to. This method should only be used for tests to make sure the recording is
+   * being used properly.
+   */
+  @VisibleForTesting
+  public Recording getRecording() {
+    synchronized (startStopLock) {
+      return recording;
+    }
+  }
```
JLLeitschuh: I don't think that synchronized provides any value here.

api/src/main/java/edu/wpi/first/shuffleboard/api/sources/recording/Recording.java
```diff
@@ -33,6 +34,26 @@ public void append(TimestampedData data) {
...
+    }
+  }
+
+  /**
+   * Gets the markers in this recording.
+   */
+  public List<Marker> getMarkers() {
+    synchronized (modificationLock) {
+      return markers;
+    }
```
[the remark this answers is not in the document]
JLLeitschuh: How is this resolved? There's no value added by having the `synchronized` here.
You aren't modifying anything or reading anything that isn't already `final` in this method.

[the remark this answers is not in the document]
JLLeitschuh: `Collections.unmodifiableList(markers)` doesn't help you here. If a thread calls `add` while another thread is iterating over the elements returned by this collection, you'll get a concurrent modification exception.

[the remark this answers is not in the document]
JLLeitschuh: Also, consider using a read/write lock.

build.gradle.kts
```diff
@@ -134,6 +134,7 @@ allprojects {
         "testCompile"(junitJupiter(name = "junit-jupiter-api"))
         "testCompile"(junitJupiter(name = "junit-jupiter-engine"))
         "testCompile"(junitJupiter(name = "junit-jupiter-params"))
+        "testCompile"(create(group = "org.junit-pioneer", name = "junit-pioneer", version = "0.3.0"))
```
JLLeitschuh:
```suggestion
        "testCompile"(group = "org.junit-pioneer", name = "junit-pioneer", version = "0.3.0")
```

build.gradle.kts
```diff
@@ -134,6 +134,7 @@ allprojects {
...
         "testRuntime"(create(group = "org.junit.platform", name = "junit-platform-launcher", version = "1.0.0"))
```
JLLeitschuh:
```suggestion
        "testRuntime"(group = "org.junit.platform", name = "junit-platform-launcher", version = "1.0.0")
```

api/src/main/java/edu/wpi/first/shuffleboard/api/sources/recording/Marker.java
```diff
@@ -0,0 +1,94 @@
...
+  /**
+   * Creates a new event marker with no description.
+   *
+   * @param name       the name of the marked event
+   * @param importance the importance of the event
+   * @param timestamp  the time the event occurred, measured in milliseconds since the start of the recording
+   */
+  public Marker(String name, MarkerImportance importance, long timestamp) {
+    this(name, "", importance, timestamp);
+  }
```
JLLeitschuh: I'm going to reiterate what I've suggested before, it's always better to not expose constructors directly and instead expose static factory methods.

api/src/main/java/edu/wpi/first/shuffleboard/api/sources/recording/Recording.java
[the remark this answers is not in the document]
SamCarlberg: The synchronization here prevents concurrent modification because `addMarker` also uses the same lock.

Read/write locks are really only useful if there are many reader threads. Though I think a lock would be better than the current implementation.

api/src/main/java/edu/wpi/first/shuffleboard/api/sources/recording/Recording.java
```diff
@@ -30,53 +41,110 @@ public void append(TimestampedData data) {
...
+   * <p>In live recordings, this list will not be exhaustive. Data is removed after saving to disk to reduce
+   * memory and CPU usage. Recordings loaded in playback mode will have all the data present.
+   *
+   * @return the data in this recording
+   */
   public List<TimestampedData> getData() {
-    synchronized (modificationLock) {
-      return data;
-    }
+    return ThreadUtils.withLock(lock.readLock(), () -> dataReadOnly);
```
JLLeitschuh: Possible concurrent modification exception here when iterating and a different thread calls `clear`.

api/src/main/java/edu/wpi/first/shuffleboard/api/sources/recording/Recording.java
```diff
@@ -30,53 +41,110 @@ public void append(TimestampedData data) {
...
+   * use {@link #copyData copyData()} to avoid locking issues (the individual methods each lock, but the lists may
+   * change between method calls).
+   *
+   * <p>In live recordings, this list will not be exhaustive. Markers are removed after saving to disk to reduce
+   * memory and CPU usage. Recordings loaded in playback mode will have all the markers present.
+   *
+   * @return the markers in this recording
+   */
+  public List<Marker> getMarkers() {
+    return ThreadUtils.withLock(lock.readLock(), () -> markersReadOnly);
```
JLLeitschuh: Possible concurrent modification exception here when iterating and a different thread calls `clear`.

api/src/main/java/edu/wpi/first/shuffleboard/api/sources/recording/Recording.java
```diff
@@ -30,77 +38,132 @@ public void append(TimestampedData data) {
...
+    private final ImmutableList<TimestampedData> data;
+    private final ImmutableList<Marker> markers;
 
-      return this.data.equals(that.data);
+    @SuppressWarnings("PMD.DefaultPackage") // This should only be used by the Serialization class in the same package
+    static Snapshot of(Recording recording) {
+      return new Snapshot(
+          ImmutableList.copyOf(recording.data),
+          ImmutableList.copyOf(recording.markers)
+      );
```
JLLeitschuh: I'd recommend indicating that this method itself isn't thread-safe. I know you're calling it in a thread safe way, but this method alone isn't.
There's also no harm in making it thread safe by calling this constructor in a `ThreadUtils.withLock(recording.lock.readLock(), () -> ....` block.
You can graph the read lock from within the write lock just FYI.

api/src/main/java/edu/wpi/first/shuffleboard/api/sources/recording/Serialization.java
```diff
@@ -110,13 +142,28 @@ private Serialization() {
   public static void saveRecording(Recording recording, Path file) throws IOException {
     Serializers.getAdapters().forEach(a -> a.setCurrentFile(file.toFile()));
     // Work on a copy, since the recording can have new data added to it while we're in the middle of saving
-    final List<TimestampedData> dataCopy = new ArrayList<>(recording.getData());
-    recording.getData().clear();
+    var snapshot = recording.takeSnapshotAndClear();
+    final var dataCopy = new ArrayList<>(snapshot.getData());
```
JLLeitschuh: Why copy it again?

SamCarlberg: Because it gets sorted, which `ImmutbaleList` doesn't support

SamCarlberg: I suppose an alternative would be
```java
var dataCopy = snapshot.getData().stream().sorted().collect(ImmutableList.toImmutableList());
```

api/src/main/java/edu/wpi/first/shuffleboard/api/util/ThreadUtils.java
```diff
@@ -27,4 +29,36 @@ public static Thread makeDaemonThread(Runnable runnable) {
...
+   * @return the output of {@code function}
+   */
+  public static <T> T withLock(Lock lock, Supplier<? extends T> function) {
+    try {
+      lock.lock();
+      return function.get();
+    } finally {
+      lock.unlock();
+    }
+  }
```
JLLeitschuh: You could also create a decorator over `Lock` that adds these methods to the interfaces as well. Just a thought.

This is the interface:
```java
public interface ReadWriteLock {
    Lock readLock();

    Lock writeLock();
}
```

Just create your own implementation:

```java
public interface FunctionalReadWriteLock extends ReaWriteLock {
   @Override
   FunctionalLock readLock();
   @Override
   FunctionalLock writeLock();
}
```
Ect...
Or you could do something like this:
https://github.com/debezium/debezium/blob/master/debezium-core/src/main/java/io/debezium/util/FunctionalReadWriteLock.java

api/src/main/java/edu/wpi/first/shuffleboard/api/sources/recording/Recording.java
SamCarlberg: I don't think this needs to be explicitly threadsafe. It'll also trigger PMD or spotbugs about referencing a private field in a parent class

JLLeitschuh: > It'll also trigger PMD or spotbugs about referencing a private field in a parent class
You're already doing that man...

PR #540, opened 2018-11-06
Remove Travis and Appveyor integration

SamCarlberg:
Not much point in using Travis and Appveyor when Azure pipelines is already running jobs for all platforms


JLLeitschuh:
All that fun Docker logic blown away 😢. Totally understandable though.
I'm glad it served you well for the year or so that it ran for.

PR #545, opened 2018-11-18
Only attempt to save a frame from a camera stream if recording

### LINKED ISSUE #544: Camera Widget leaks memory
**Describe the bug**  
Creating a camera widget causes memory usage to rapidly increase, making the system unusable after a few minutes.

**To Reproduce**  
Connect to a camera stream and add a camera widget.  Then monitor shuffleboard's memory usage.

**Expected behavior**  
For shuffleboard to maintain relatively constant memory usage.

 - OS: Ubuntu 18.04 on x86-64
 - Java version: OpenJDK 11 (build 11+24-Ubuntu-118.04)
 - Version: 2019.1.1-beta-2.

**Additional context**  
Commenting out [this line](https://github.com/wpilibsuite/shuffleboard/blob/6016fab3581889cb4de1aff57548086c01b8ca14/plugins/cameraserver/src/main/java/edu/wpi/first/shuffleboard/plugin/cameraserver/source/CameraServerSource.java#L254) makes the issue go away.    I suspect this is similar to #438.  The `Mat` gets cloned once per frame but is never freed, but since the memory is managed by native code Java doesn't think it needs to garbage collect.

SamCarlberg:
Fixes a rather large potential memory leak caused by keeping camera frames in memory waiting to be saved to disk but never are. 

Fixes #544

### DIFF
--- a/plugins/cameraserver/src/main/java/edu/wpi/first/shuffleboard/plugin/cameraserver/source/CameraServerSource.java
+++ b/plugins/cameraserver/src/main/java/edu/wpi/first/shuffleboard/plugin/cameraserver/source/CameraServerSource.java
@@ -251,7 +251,9 @@ public final class CameraServerSource extends AbstractDataSource<CameraServerDat
       } else {
         setData(getData().withImage(image));
       }
-      Recorder.getInstance().record(getId(), getDataType(), getData().withImage(image.clone()));
+      if (Recorder.getInstance().isRunning()) {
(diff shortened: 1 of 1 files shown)


blucoat:
If recording is turned on, would this still be a memory leak, since the recording code doesn't free the `Mat`s after writing them to disc?

SamCarlberg:
No. The image is freed after it is written to disk in the class that handles saving the images to the .mp4 file.

blucoat:
Welp, just noticed that after commenting.  Thanks for the speedy fix!

PR #554, opened 2018-12-11
Explicitly stop the ntcore client on shutdown

SamCarlberg:
Fixes a deadlock in the ntcore shutdown sequence (probably from https://github.com/wpilibsuite/allwpilib/pull/1357)

Note that the deadlock only seems to occur on Windows 7; tests on Linux and Windows 10 have had clean shutdowns


ThadHouse:
Fix merged upstream.

PR #557, opened 2018-12-12
Allow complex widget property types to be specified remotely

SamCarlberg:
# Overview

This allows custom widget properties to be properly parsed and set - e.g. setting custom colors on a boolean box widget from a robot program.

Parsers are available by default for numbers, strings, booleans, JavaFX colors, orientation, and layout label positions. The base and cameraserver plugins also have parsers for custom widget properties in those plugins.

### DIFF
--- a/api/src/main/java/edu/wpi/first/shuffleboard/api/PropertyParser.java
+++ b/api/src/main/java/edu/wpi/first/shuffleboard/api/PropertyParser.java
@@ -1,0 +1,80 @@
+package edu.wpi.first.shuffleboard.api;
+
+/**
+ * Parses arbitrary input and returns the result. This is used to set the values of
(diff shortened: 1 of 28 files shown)


api/src/main/java/edu/wpi/first/shuffleboard/api/PropertyParser.java
```diff
@@ -0,0 +1,80 @@
...
+        }
+        E[] values = type.getEnumConstants();
+        if (input instanceof String) {
+          for (E value : values) {
+            if (value.name().equals(input)) {
+              return value;
+            }
+          }
+        } else if (input instanceof Integer) {
+          return values[(Integer) input];
```
JLLeitschuh: Uses ordinal?

PR #568, opened 2019-01-01
Ultrasonic widget

SamCarlberg:
# Overview
- Adds a new widget for displaying ultrasonic sensors
Assumes the given input is in inches (that's what the WPILib objects give)
The display unit can be changed to inches, feet, meters, centimeters, or millimeters, and will convert the input value from inches to the selected unit

- Fixes the plugin loader not applying custom property parsers

# Screenshots
### Ultrasonic widgets generated from a robot program
image


andrewda:
Could this be generalized more to a "Distance Widget"? I could see that being used for ultrasonic sensors, lidar sensors, lift height, distance driven, etc. and it'd be very useful to convert those values to a different unit of measurement on the dashboard.

SamCarlberg:
A generic widget for displaying a distance would be nice. It'll need a lot of design work to be feasible.

For now, a simple widget for ultrasonic sensors is sufficient.

PR #569, opened 2019-01-03
Fix uncaught exceptions not ending up in the log file

JLLeitschuh:
# Overview
Makes it so that errors that are caused at launch also end up in the log file. Eg. The issue @bradamiller had this evening.

### DIFF
--- a/app/src/main/java/edu/wpi/first/shuffleboard/app/Main.java
+++ b/app/src/main/java/edu/wpi/first/shuffleboard/app/Main.java
@@ -2,6 +2,9 @@ package edu.wpi.first.shuffleboard.app;
 
 import com.sun.javafx.application.LauncherImpl;
 
+import java.util.logging.Level;
+import java.util.logging.Logger;
+
 /**
  * The true main class.  This bypasses module boot layer introspection by the Java launcher that attempts to
  * reflectively access the JavaFX application launcher classes - this will fail because there is no module path;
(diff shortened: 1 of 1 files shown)


app/src/main/java/edu/wpi/first/shuffleboard/app/Main.java
```diff
@@ -10,11 +13,22 @@
...
   public static void main(String[] args) {
     // JavaFX 11+ uses GTK3 by default, and has problems on some display servers
     // This flag forces JavaFX to use GTK2
     System.setProperty("jdk.gtk.version", "2");
+
+    Thread
+        .setDefaultUncaughtExceptionHandler((thread, throwable) -> {
+          logger.log(
+              Level.SEVERE,
+              "Uncaught exception on thread `" + thread.getName() + "`.",
```
SamCarlberg: Why the backticks here?

app/src/main/java/edu/wpi/first/shuffleboard/app/Main.java
```diff
@@ -10,11 +13,22 @@
...
+  private static final Logger logger = Logger.getLogger(Main.class.getName());
+
   @SuppressWarnings("JavadocMethod")
   public static void main(String[] args) {
     // JavaFX 11+ uses GTK3 by default, and has problems on some display servers
     // This flag forces JavaFX to use GTK2
     System.setProperty("jdk.gtk.version", "2");
+
+    Thread
+        .setDefaultUncaughtExceptionHandler((thread, throwable) -> {
```
SamCarlberg: I'd use a method reference instead of a lambda, just to make the log a bit more descriptive.

For reference, this is what logs would look like:
```
Jan 06, 2019 4:46:00 PM edu.wpi.first.shuffleboard.app.Main lambda$main$0
```

app/src/main/java/edu/wpi/first/shuffleboard/app/Main.java
SamCarlberg: Also, this is always on the main thread so there's no need to parameterize it in the message.

JLLeitschuh: You don't ever spawn other threads?

SamCarlberg: We do, but this handler is only used for the main thread

PR #571, opened 2019-01-08
Improve Camera Widget

SamCarlberg:
# Overview

Some minor improvements for camera widgets:
- Use a [JFoenix](https://github.com/jfoenixadmin/JFoenix) slider for compression to show the actual value (instead of an inscrutable slider)
- Fix loading of camera widget properties from save file
- Allow compression to not be set with a value of `-1` (slider in the farthest left position) to reduce CPU usage on the stream host

# Screenshots

screen record from 2019-01-08 13 36 08

### DIFF
--- a/plugins/cameraserver/src/main/java/edu/wpi/first/shuffleboard/plugin/cameraserver/widget/CameraServerWidget.java
+++ b/plugins/cameraserver/src/main/java/edu/wpi/first/shuffleboard/plugin/cameraserver/widget/CameraServerWidget.java
@@ -162,9 +162,9 @@ public class CameraServerWidget extends SimpleAnnotatedWidget<CameraServerData> 
     if (getSource() instanceof CameraServerSource) {
       CameraServerSource source = (CameraServerSource) getSource();
       int compression = (int) compressionSlider.getValue();
-      int fps = frameRateField.getNumber();
-      int width = this.width.getNumber();
-      int height = this.height.getNumber();
+      int fps = ((Number) frameRateField.getNumber()).intValue();
+      int width = ((Number) this.width.getNumber()).intValue();
+      int height = ((Number) this.height.getNumber()).intValue();
       boolean change = source.getTargetCompression() != compression
           || source.getTargetFps() != fps
           || source.getTargetResolution().isNotEqual(width, height);
--- a/plugins/cameraserver/src/main/java/edu/wpi/first/shuffleboard/plugin/cameraserver/widget/CameraServerWidget.java
+++ b/plugins/cameraserver/src/main/java/edu/wpi/first/shuffleboard/plugin/cameraserver/widget/CameraServerWidget.java
@@ -13,11 +13,13 @@ import edu.wpi.first.shuffleboard.plugin.cameraserver.recording.serialization.Im
 import edu.wpi.first.shuffleboard.plugin.cameraserver.source.CameraServerSource;
 
 import com.google.common.collect.ImmutableList;
+import com.jfoenix.controls.JFXSlider;
 
-import java.util.List;
 import org.fxmisc.easybind.EasyBind;
 import org.opencv.core.Core;
 import org.opencv.core.Mat;
+
+import java.util.List;
 
 import javafx.beans.property.BooleanProperty;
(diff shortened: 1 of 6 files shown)


JLLeitschuh:
What are the units on "compression"? You might want to include that.

SamCarlberg:
Compression is weird ... it's more of a quality level, where 0 is lowest quality (and highest compression ratio) and 100 is highest quality. But there's no real unit used to describe it

JLLeitschuh:
> But there's no real unit used to describe it

Use a percentage then.

ThadHouse:
Its not a percentage though. Its not linear, and -1 is also a valid value.

plugins/cameraserver/src/main/java/edu/wpi/first/shuffleboard/plugin/cameraserver/widget/CameraServerWidget.java
```diff
@@ -161,9 +162,9 @@ private void applySettings() {
     if (getSource() instanceof CameraServerSource) {
       CameraServerSource source = (CameraServerSource) getSource();
       int compression = (int) compressionSlider.getValue();
-      int fps = frameRateField.getNumber();
-      int width = this.width.getNumber();
-      int height = this.height.getNumber();
+      int fps = ((Number) frameRateField.getNumber()).intValue();
+      int width = ((Number) this.width.getNumber()).intValue();
+      int height = ((Number) this.height.getNumber()).intValue();
```
JLLeitschuh: Why is this now necessary?
The types haven't changed, have they?

plugins/cameraserver/src/main/resources/edu/wpi/first/shuffleboard/plugin/cameraserver/widget/CameraServerWidget.fxml
```diff
@@ -14,6 +14,7 @@
 <?import javafx.scene.layout.GridPane?>
 <?import javafx.scene.layout.Pane?>
 <?import javafx.scene.control.Button?>
+<?import com.jfoenix.controls.JFXSlider?>
```
JLLeitschuh: You have a now unused import a few lines up: `<?import javafx.scene.control.Slider?>`

JLLeitschuh:
> Its not a percentage though. Its not linear, and -1 is also a valid value.

Looking at the graphic above, the min and max values are 0-100. It doesn't look like `-1` is a supported input with the current API.

plugins/cameraserver/src/main/java/edu/wpi/first/shuffleboard/plugin/cameraserver/widget/CameraServerWidget.java
SamCarlberg: Gson deserializes numbers are their own special type, which are then loaded into these properties (which cannot be checked at runtime due to erasure). Controls on a camera widget that was in a save file cannot be used because Java's trying to unbox an `Integer` into an `int`, but the runtime type may be a `LazilyParsedNumber` instead

SamCarlberg:
I changed the minimum value to `-1` in 717e146. Didn't update the graphic since it's a pain to set up the recording

plugins/cameraserver/src/main/java/edu/wpi/first/shuffleboard/plugin/cameraserver/widget/CameraServerWidget.java
JLLeitschuh: Derp. Wow. But why are you casing the type of `this.height.getNumber` which is of type `Number` to `Number`?

SamCarlberg: It's of type `Integer` (compile time) but can possibly be a different subclass of `Number` at runtime

SamCarlberg: Reverted this change, since the underlying issue was fixed in #572

PR #591, opened 2019-02-28
Allow users to manually specify upper and lower bounds on graph Y-axis

### LINKED ISSUE #589: Add min and max y-axis value to properties for Shuffleboard Graph widget
Currently, we can only specify the size of the x-axis.  It'd be really nice to be able to control the y-axis, as well.

SamCarlberg:
# Overview

Adds three new settings to the graph widget's Y-axis:

### Automatic bounds
Use the default JavaFX graph behavior of determining the upper and lower bounds for a graph' Y-axis

### Upper bound
Sets the upper bound of the Y-axis. Defaults to `1`. Can only be used if `Automatic bounds` is disabled

### Lower bound
Sets the lower bound of the Y-axis. Defaults to `-1`. Can only be used if `Automatic bounds` is disabled

Closes #589 
# Screenshots

screenshot from 2019-02-27 23 15 59


Oblarg:
These can be set from the WPILib Shuffleboard API, correct?

PR #593, opened 2019-03-02
Support LabVIEW cameras

ThadHouse:
Their URL's are invalid, so remove the one they pass, and add the camera name.

Requires https://github.com/wpilibsuite/allwpilib/pull/1621

### DIFF
--- a/plugins/cameraserver/src/main/java/edu/wpi/first/shuffleboard/plugin/cameraserver/source/CameraUrlGenerator.java
+++ b/plugins/cameraserver/src/main/java/edu/wpi/first/shuffleboard/plugin/cameraserver/source/CameraUrlGenerator.java
@@ -91,7 +91,7 @@ public final class CameraUrlGenerator {
 
     if (query != null) {
       // Handle the NI special case
-      String niName = commands.getOrDefault("name", null);
+      String niName = commands.get("name");
 
       String[] existingCommands = query.split("&");
       for (String command : existingCommands) {
@@ -130,7 +130,7 @@ public final class CameraUrlGenerator {
   }
 
   private static String httpUrlEncode(Map.Entry<String, String> rawCommand) {
-    return URLEncoder.encode(rawCommand.getKey(), StandardCharsets.UTF_8).replaceAll("\\+", "%20") + "="
-        + URLEncoder.encode(rawCommand.getValue(), StandardCharsets.UTF_8).replaceAll("\\+", "%20") ;
+    return URLEncoder.encode(rawCommand.getKey(), StandardCharsets.UTF_8).replace("+", "%20") + "="
+        + URLEncoder.encode(rawCommand.getValue(), StandardCharsets.UTF_8).replace("+", "%20") ;
   }
 }
--- a/plugins/cameraserver/src/main/java/edu/wpi/first/shuffleboard/plugin/cameraserver/source/CameraUrlGenerator.java
+++ b/plugins/cameraserver/src/main/java/edu/wpi/first/shuffleboard/plugin/cameraserver/source/CameraUrlGenerator.java
@@ -93,10 +93,10 @@ public final class CameraUrlGenerator {
 
   private static String httpUrlEncode(Map.Entry<String, String> rawCommand) {
     try {
-      return URLEncoder.encode(e.getKey(), "utf-8").replaceAll("\\+", "%20") + "="
-          + URLEncoder.encode(e.getValue(), "utf-8").replaceAll("\\+", "%20") ;
+      return URLEncoder.encode(rawCommand.getKey(), "utf-8").replaceAll("\\+", "%20") + "="
+          + URLEncoder.encode(rawCommand.getValue(), "utf-8").replaceAll("\\+", "%20") ;
     } catch (UnsupportedEncodingException ex) {
-      return e.getKey() + "=" + e.getValue();
+      return rawCommand.getKey() + "=" + rawCommand.getValue();
     }
   }
 
--- a/plugins/cameraserver/src/main/java/edu/wpi/first/shuffleboard/plugin/cameraserver/source/CameraUrlGenerator.java
+++ b/plugins/cameraserver/src/main/java/edu/wpi/first/shuffleboard/plugin/cameraserver/source/CameraUrlGenerator.java
@@ -47,7 +47,9 @@ public final class CameraUrlGenerator {
     if (frameRate > 0) {
       commands.put("fps", Integer.toString(frameRate));
     }
-    commands.put("name", source.getName());
+    if (source.getName().contains("IMAQdx")) {
+      commands.put("name", source.getName());
+    }
     return generateUrls(commands, baseUrls, source.getName());
   }
 
--- a/plugins/cameraserver/src/main/java/edu/wpi/first/shuffleboard/plugin/cameraserver/source/CameraUrlGenerator.java
+++ b/plugins/cameraserver/src/main/java/edu/wpi/first/shuffleboard/plugin/cameraserver/source/CameraUrlGenerator.java
@@ -13,6 +13,7 @@ import java.io.UnsupportedEncodingException;
 import java.net.MalformedURLException;
 import java.net.URI;
 import java.net.URISyntaxException;
+import java.net.URLDecoder;
 import java.net.URLEncoder;
 
 /**
--- a/plugins/cameraserver/src/main/java/edu/wpi/first/shuffleboard/plugin/cameraserver/source/CameraUrlGenerator.java
+++ b/plugins/cameraserver/src/main/java/edu/wpi/first/shuffleboard/plugin/cameraserver/source/CameraUrlGenerator.java
@@ -93,13 +93,17 @@ public final class CameraUrlGenerator {
       // Handle the NI special case
       String niName = commands.getOrDefault("name", null);
 
-      String[] existingCommands = uri.getQuery().split("&");
+      String[] existingCommands = query.split("&");
       for (String command : existingCommands) {
         String[] commandSplit = command.split("=");
         if (commandSplit.length != 2) {
           continue;
         }
-        commands.put(commandSplit[0], commandSplit[1]);
+        try {
+          commands.put(URLDecoder.decode(commandSplit[0], "utf-8"), URLDecoder.decode(commandSplit[1], "utf-8"));
+        } catch (UnsupportedEncodingException e1) {
+          commands.put(commandSplit[0], commandSplit[1]);
+        }
       }
       if (niName != null) {
         commands.put("name", niName);
@@ -109,12 +113,24 @@ public final class CameraUrlGenerator {
     var queryStr = commands.entrySet().stream()
         .map(e -> httpUrlEncode(e))
         .collect(Collectors.joining("&"));
-    try {
-      return new URI(uri.getScheme(), uri.getAuthority(), uri.getPath(), queryStr,
-          uri.getFragment()).toString();
-    } catch (URISyntaxException e1) {
-      return input;
+      return encodeUri(uri, queryStr);
+
+  }
+
+  private static String encodeUri(URI uri, String queryStr) {
+    StringBuilder builder = new StringBuilder();
+    builder.append(uri.getScheme());
+    builder.append("://");
+    builder.append(uri.getAuthority());
+    builder.append(uri.getPath());
+    builder.append('?');
+    builder.append(queryStr);
+    String fragment = uri.getFragment();
+    if (fragment != null) {
+      builder.append('#');
+      builder.append(fragment);
     }
+    return builder.toString();
   }
 
   private static String httpUrlEncode(Map.Entry<String, String> rawCommand) {
--- a/plugins/cameraserver/src/main/java/edu/wpi/first/shuffleboard/plugin/cameraserver/source/CameraUrlGenerator.java
+++ b/plugins/cameraserver/src/main/java/edu/wpi/first/shuffleboard/plugin/cameraserver/source/CameraUrlGenerator.java
@@ -33,8 +33,7 @@ public final class CameraUrlGenerator {
    *
    * @param baseUrls the base stream URLs
    */
-  public String[] generateUrls(String[] baseUrls,
-                               String cameraName) { // NOPMD varargs instead of array
+  public String[] generateUrls(String[] baseUrls) { // NOPMD varargs instead of array
     Map<String, String> commands = new LinkedHashMap<>();
     Resolution resolution = source.getTargetResolution();
     if (resolution != null && resolution.getWidth() > 0 && resolution.getHeight() > 0) {
@@ -48,8 +47,8 @@ public final class CameraUrlGenerator {
     if (frameRate > 0) {
       commands.put("fps", Integer.toString(frameRate));
     }
-    commands.put("name", cameraName);
-    return generateUrls(commands, baseUrls, cameraName);
+    commands.put("name", source.getName());
+    return generateUrls(commands, baseUrls, source.getName());
   }
 
   @VisibleForTesting
@@ -83,7 +82,8 @@ public final class CameraUrlGenerator {
     var commandStr = commands.entrySet().stream()
         .map(e -> {
           try {
-            return URLEncoder.encode(e.getKey(), "utf-8").replaceAll("\\+", "%20") + "=" + URLEncoder.encode(e.getValue(), "utf-8").replaceAll("\\+", "%20") ;
+            return URLEncoder.encode(e.getKey(), "utf-8").replaceAll("\\+", "%20") + "="
+                + URLEncoder.encode(e.getValue(), "utf-8").replaceAll("\\+", "%20") ;
           } catch (UnsupportedEncodingException ex) {
             return e.getKey() + "=" + e.getValue();
           }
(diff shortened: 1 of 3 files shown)


plugins/cameraserver/src/main/java/edu/wpi/first/shuffleboard/plugin/cameraserver/source/CameraUrlGenerator.java
```diff
@@ -45,31 +47,53 @@ public CameraUrlGenerator(CameraServerSource source) {
...
     }
-    return generateUrls(commands, baseUrls);
+    commands.put("name", source.getName());
+    return generateUrls(commands, baseUrls, source.getName());
   }
 
   @VisibleForTesting
-  static String[] generateUrls(Map<String, String> commands, String[] baseUrls) { // NOPMD varargs instead of array
+  static String[] generateUrls(Map<String, String> commands, String[] baseUrls,
+                               String cameraName) { // NOPMD varargs instead of array
```
SamCarlberg: Don't need the NOPMD comment if `baseUrls` is no longer the final method parameter

ThadHouse: Fixed

plugins/cameraserver/src/main/java/edu/wpi/first/shuffleboard/plugin/cameraserver/source/CameraUrlGenerator.java
```diff
@@ -45,31 +47,53 @@ public CameraUrlGenerator(CameraServerSource source) {
...
-    return commands.entrySet().stream()
-        .map(e -> e.getKey() + "=" + e.getValue())
+    // Special case to remove name from LabVIEW camera
+    if (commands.containsKey("name") && commands.get("name").contains("IMAQdx")) {
+      input = input.replaceAll("\\?name=cam\\d", "");
+    }
+    var commandStr = commands.entrySet().stream()
+        .map(e -> {
+          try {
+            return URLEncoder.encode(e.getKey(), "utf-8").replaceAll("\\+", "%20") + "="
```
SamCarlberg: I'd pull the URLEncoder stuff into its own method.

```java
static String httpEncode(String raw)
```

It'd also make this nicer since you wouldn't need the try-catch in the lambda here

ThadHouse: It can't take a string, it has to take the Map Entry, but yeah I'll do that.

plugins/cameraserver/src/main/java/edu/wpi/first/shuffleboard/plugin/cameraserver/source/CameraUrlGenerator.java
```diff
@@ -45,31 +47,53 @@ public CameraUrlGenerator(CameraServerSource source) {
...
+  static String toHttpParams(String input, Map<String, String> commands) {
     if (commands.isEmpty()) {
-      return "";
+      return input;
     }
-    return commands.entrySet().stream()
-        .map(e -> e.getKey() + "=" + e.getValue())
+    // Special case to remove name from LabVIEW camera
+    if (commands.containsKey("name") && commands.get("name").contains("IMAQdx")) {
+      input = input.replaceAll("\\?name=cam\\d", "");
```
SamCarlberg: Would it be possible to check for `IMAQdx` in the `generateUrls` method when adding the `name` parameter? It seems like that'd be simpler to do than the current approach of unconditionally adding it there, then removing it in this method.

plugins/cameraserver/src/main/java/edu/wpi/first/shuffleboard/plugin/cameraserver/source/CameraUrlGenerator.java
SamCarlberg: I was envisioning something like `return httpEncode(e.getKey()) + "=" + httpEncode(e.getValue())`, but that'd work too

plugins/cameraserver/src/main/java/edu/wpi/first/shuffleboard/plugin/cameraserver/source/CameraUrlGenerator.java
ThadHouse: We're not removing it in this method. The url that is in the NT publisher already includes `name=camx`, and cscore cannot currently support multiple named parameters. So this removed the one already included, not the extra one that would be added. CS Core might support name encoding in the future, and we would use the name like labview likely, so thats why the new name parameter is added unconditionally.

plugins/cameraserver/src/main/java/edu/wpi/first/shuffleboard/plugin/cameraserver/source/CameraUrlGenerator.java
ThadHouse: I can't do that, because otherwise that `=` gets encoded by the `encode` method. Which we cannot have.

plugins/cameraserver/src/main/java/edu/wpi/first/shuffleboard/plugin/cameraserver/source/CameraUrlGenerator.java
ThadHouse: Actually, when we add it we wouldn't need the extra name, so I'll conditional that too.

ThadHouse: I conditionally added it on the other end as well, as otherwise that name could interfere with a stream that validly uses name.

ThadHouse: I conditionally added it on the other end as well, as otherwise that name could interfere with a stream that validly uses name.

plugins/cameraserver/src/main/java/edu/wpi/first/shuffleboard/plugin/cameraserver/source/CameraUrlGenerator.java
```diff
@@ -45,31 +51,95 @@ public CameraUrlGenerator(CameraServerSource source) {
...
+      builder.append('#');
+      builder.append(fragment);
+    }
+    return builder.toString();
   }
 
+  private static String httpUrlEncode(Map.Entry<String, String> rawCommand) {
+    try {
+      return URLEncoder.encode(rawCommand.getKey(), "utf-8").replaceAll("\\+", "%20") + "="
+          + URLEncoder.encode(rawCommand.getValue(), "utf-8").replaceAll("\\+", "%20") ;
```
SamCarlberg: If you're just replacing the `+` character, use `replace("+", "%20")`. Faster and more obvious than a regular expression

plugins/cameraserver/src/main/java/edu/wpi/first/shuffleboard/plugin/cameraserver/source/CameraUrlGenerator.java
```diff
@@ -45,31 +51,95 @@ public CameraUrlGenerator(CameraServerSource source) {
...
+      String niName = commands.getOrDefault("name", null);
+
+      String[] existingCommands = query.split("&");
+      for (String command : existingCommands) {
+        String[] commandSplit = command.split("=");
+        if (commandSplit.length != 2) {
+          continue;
+        }
+        try {
+          commands.put(URLDecoder.decode(commandSplit[0], "utf-8"), URLDecoder.decode(commandSplit[1], "utf-8"));
```
SamCarlberg: You can use `StandardCharsets.UTF_8` instead of the string `"utf-8"`. That overload also does not throw a checked exception

plugins/cameraserver/src/main/java/edu/wpi/first/shuffleboard/plugin/cameraserver/source/CameraUrlGenerator.java
```diff
@@ -45,31 +51,95 @@ public CameraUrlGenerator(CameraServerSource source) {
...
+    if (fragment != null) {
+      builder.append('#');
+      builder.append(fragment);
+    }
+    return builder.toString();
   }
 
+  private static String httpUrlEncode(Map.Entry<String, String> rawCommand) {
+    try {
+      return URLEncoder.encode(rawCommand.getKey(), "utf-8").replaceAll("\\+", "%20") + "="
```
SamCarlberg: Can use `StandardCharsets.UTF_8` here

plugins/cameraserver/src/main/java/edu/wpi/first/shuffleboard/plugin/cameraserver/source/CameraUrlGenerator.java
```diff
@@ -45,31 +51,95 @@ public CameraUrlGenerator(CameraServerSource source) {
...
+      return input;
     }
-    return commands.entrySet().stream()
-        .map(e -> e.getKey() + "=" + e.getValue())
+
+    String query = uri.getQuery();
+
+    if (query != null) {
+      // Handle the NI special case
+      String niName = commands.getOrDefault("name", null);
```
SamCarlberg: Can just use `commands.get("name")`, since that will return `null` if the key is not in the map

plugins/cameraserver/src/main/java/edu/wpi/first/shuffleboard/plugin/cameraserver/source/CameraUrlGenerator.java
```diff
@@ -45,31 +51,95 @@ public CameraUrlGenerator(CameraServerSource source) {
...
+                               String cameraName) {
     if (baseUrls == null || baseUrls.length == 0) {
       return new String[0];
     }
     if (commands.isEmpty()) {
       return baseUrls;
     } else {
-      return Arrays.stream(baseUrls)
-          .map(url -> url + toHttpParams(commands))
+      var urls =  Arrays.stream(baseUrls)
```
SamCarlberg: Don't need the intermediate variable, just return the result directly

plugins/cameraserver/src/main/java/edu/wpi/first/shuffleboard/plugin/cameraserver/source/CameraUrlGenerator.java
```diff
@@ -45,31 +51,95 @@ public CameraUrlGenerator(CameraServerSource source) {
...
+          .map(url -> addHttpParams(url, commands))
+          .filter(url -> url != null)
```
SamCarlberg: Use the `Objects::nonNull` method reference

plugins/cameraserver/src/main/java/edu/wpi/first/shuffleboard/plugin/cameraserver/source/CameraUrlGenerator.java
```diff
@@ -45,31 +51,95 @@ public CameraUrlGenerator(CameraServerSource source) {
...
+    if (source.getName().contains("IMAQdx")) {
+      commands.put("name", source.getName());
+    }
+    return generateUrls(commands, baseUrls, source.getName());
   }
 
   @VisibleForTesting
-  static String[] generateUrls(Map<String, String> commands, String[] baseUrls) { // NOPMD varargs instead of array
+  static String[] generateUrls(Map<String, String> commands, String[] baseUrls,
+                               String cameraName) {
```
SamCarlberg: The `cameraName` parameter is no longer used in the method body and should be removed. The NOPMD comment doesn't need to be re-added, since we've updated our PMD ruleset since this method was originally written.

plugins/cameraserver/src/main/java/edu/wpi/first/shuffleboard/plugin/cameraserver/source/CameraUrlGenerator.java
```diff
@@ -45,31 +51,95 @@ public CameraUrlGenerator(CameraServerSource source) {
...
+    } catch (URISyntaxException ex) {
+      return input;
     }
-    return commands.entrySet().stream()
-        .map(e -> e.getKey() + "=" + e.getValue())
+
+    String query = uri.getQuery();
+
+    if (query != null) {
+      // Handle the NI special case
```
SamCarlberg: Please add unit tests for NI camera URLs

plugins/cameraserver/src/main/java/edu/wpi/first/shuffleboard/plugin/cameraserver/source/CameraUrlGenerator.java
```diff
@@ -45,31 +51,95 @@ public CameraUrlGenerator(CameraServerSource source) {
...
+          commands.put(commandSplit[0], commandSplit[1]);
+        }
+      }
+      if (niName != null) {
+        commands.put("name", niName);
+      }
+    }
+
+    var queryStr = commands.entrySet().stream()
+        .map(e -> httpUrlEncode(e))
```
SamCarlberg: I suggest changing this lambda to a method reference  `CameraUrlGenerator::httpUrlEncode`

plugins/cameraserver/src/main/java/edu/wpi/first/shuffleboard/plugin/cameraserver/source/CameraUrlGenerator.java
ThadHouse: There can be multiple `+` characters, so `replace` doesn't work.

plugins/cameraserver/src/main/java/edu/wpi/first/shuffleboard/plugin/cameraserver/source/CameraUrlGenerator.java
ThadHouse: That doesn't throw? I thought it would.

plugins/cameraserver/src/main/java/edu/wpi/first/shuffleboard/plugin/cameraserver/source/CameraUrlGenerator.java
SamCarlberg: `String.replace` replaces all instances of the target string: https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/lang/String.html#replace(java.lang.CharSequence,java.lang.CharSequence)

plugins/cameraserver/src/main/java/edu/wpi/first/shuffleboard/plugin/cameraserver/source/CameraUrlGenerator.java
SamCarlberg: Only for map implementations that disallow null keys, like `TreeMap`

plugins/cameraserver/src/main/java/edu/wpi/first/shuffleboard/plugin/cameraserver/source/CameraUrlGenerator.java
ThadHouse: Fixed

plugins/cameraserver/src/main/java/edu/wpi/first/shuffleboard/plugin/cameraserver/source/CameraUrlGenerator.java
ThadHouse: Fixed

plugins/cameraserver/src/main/java/edu/wpi/first/shuffleboard/plugin/cameraserver/source/CameraUrlGenerator.java
ThadHouse: Fixed

plugins/cameraserver/src/main/java/edu/wpi/first/shuffleboard/plugin/cameraserver/source/CameraUrlGenerator.java
ThadHouse: Fixed

plugins/cameraserver/src/main/java/edu/wpi/first/shuffleboard/plugin/cameraserver/source/CameraUrlGenerator.java
ThadHouse: Ah ok. Fixed

plugins/cameraserver/src/main/java/edu/wpi/first/shuffleboard/plugin/cameraserver/source/CameraUrlGenerator.java
ThadHouse: Hmm I wonder where I thought it didn't replace all. I think I ran into that when working with gradle or something. If thats what it does then that works. Fixed

plugins/cameraserver/src/main/java/edu/wpi/first/shuffleboard/plugin/cameraserver/source/CameraUrlGenerator.java
SamCarlberg: Ah, I was thinking of null keys. If you call `Map.get` with a non-null key, the result will always be the value of the mapping, or `null` if nothing is mapped to the key. No map implementation will ever throw an exception if given a non-null key.

PR #605, opened 2019-04-02
Add settings to recording converters

### LINKED ISSUE #594: Add option to fill missing data in recordings to CSV converter
Currently, converted .csv shuffleboard recordings have large tracts of missing data - specifically, any time a value does not change, it is omitted.  This can make it rather difficult/annoying to process log files.  It'd be nice to have an option in the CSV converter to fill in all the missing data.

SamCarlberg:
# Overview

- Allow converters to define their own settings
- Add a CSV converter setting to fill in empty cells with the most recent value

This is a breaking change. Custom recording converters will need to be updated

Fixes #594

### DIFF
--- a/app/app.gradle.kts
+++ b/app/app.gradle.kts
@@ -31,9 +31,7 @@ dependencies {
     compile(group = "de.huxhorn.lilith", name = "de.huxhorn.lilith.3rdparty.junique", version = "1.0.4")
     compile(group = "org.apache.commons", name = "commons-csv", version = "1.5")
     testCompile(project("test_plugins"))
-    val apiTestOutput = project(":api").dependencyProject.sourceSets["test"].output
-    testCompile(files(apiTestOutput))
-    testCompile(files(apiTestOutput.resourcesDir))
+    testCompile(project(":api-test-util"))
 }
 
 val theMainClassName = "edu.wpi.first.shuffleboard.app.Main"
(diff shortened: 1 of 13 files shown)


app/app.gradle.kts
```diff
@@ -31,6 +31,9 @@ dependencies {
     compile(group = "de.huxhorn.lilith", name = "de.huxhorn.lilith.3rdparty.junique", version = "1.0.4")
     compile(group = "org.apache.commons", name = "commons-csv", version = "1.5")
     testCompile(project("test_plugins"))
+    val apiTestOutput = project(":api").dependencyProject.sourceSets["test"].output
```
JLLeitschuh: I think you can just depend upon the output of the `test` configuration for the `API` project?
Otherwise, I generally recommend creating a separate project that contains test logic you want to share.

You're allowed to have a dependency like this (-> means depends on):

project A:Test -> project ATestUtil -> project A
project A:Test -> project A

Where project A:Test is the test code for project A

SamCarlberg: Mmm, that's possible. I just grabbed this configuration logic from GRIP

JLLeitschuh: What I've suggested is cleaner.
I don't know if the task graph will generate correctly with what you have here.
It might work in most cases due to timing, but no guarantees. Maybe I'm wrong.

You'd have to look at the `compileTest` task graph with `--dry-run`.

PR #609, opened 2019-04-19
Demo of a raw cscore stream

ThadHouse:
No memory issues between models, opencv is completely removed from the wpilib requirement

Does not work, depends on the updated cscore which is not pushed yet. Just a demo of how it would.

Fill fix up the PR when that is published.

Some files are copies of whats in the new cscore (The RawFrame and RawJNI classes), but here so I could test compilation.


azure-pipelines.yml
```diff
@@ -134,10 +134,10 @@ stages:
...
         gradleOptions: '-Xmx3072m'
-        tasks: 'publish '
+        tasks: 'publish artifactoryPublish'
         options: '-PreleaseMode -PbuildServer'
       condition: and(succeeded(), startsWith(variables['Build.SourceBranch'], 'refs/tags/v'))
       env:
         RUN_AZURE_ARTIFACTORY_RELEASE: 'TRUE'
         ARTIFACTORY_PUBLISH_USERNAME: $(PublishUserName)
-        ARTIFACTORY_PUBLISH_PASSWORD: $(PublishPassword)
+        ARTIFACTORY_PUBLISH_PASSWORD: $(PublishPassword)
```
AustinShalit: EOL @ EOF?

PR #611, opened 2019-04-19
Fix in-app build times reported as the current date

SamCarlberg:
Was caused by a pretty dumb way to read the application manifest file.


PeterJohnson:
Merged as part of #633.

PR #621, opened 2019-07-24
fix syntax error in build.gradle.kts

andrewgazelka:
nullable rawUnix string and file were used when the signature has them
as non-nullable. Although it still compiles before, IntelliJ highlights
it as a critical error.

# Overview

# Screenshots

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -71,13 +71,9 @@ allprojects {
             class TrimTrailingSpaces : FormatterStep {
                 override fun getName(): String = "trimTrailingSpaces"
 
-                override fun format(rawUnix: String, file: File): String? {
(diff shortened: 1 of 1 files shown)


AustinShalit:
I am not sure this change complete's the interface contract specified by FormatterStep.

@JLLeitschuh Is this allowed or can the method still be passed a null from the Java side?

https://github.com/diffplug/spotless/blob/master/lib/src/main/java/com/diffplug/spotless/FormatterStep.java

AustinShalit:
Oh.  The interface comments specify "never-null".  Not sure we can make that guarantee though.

AustinShalit:
Nevermind re: above.

JLLeitschuh:
I don't know enough about that interface to comment, sorry.

PR #622, opened 2019-08-23
Use artifactory builds, and publish to artifactory

ThadHouse:
This ones weird on the artifactory side, but it seems to work.

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -71,7 +71,7 @@ allprojects {
                     setProperty("maven", true)
                 })
             })
-            clientConfig.info.setBuildName("ShuffleBoard")
+            clientConfig.info.setBuildName("Shuffleboard")
         }
 
         tasks.named("publish") {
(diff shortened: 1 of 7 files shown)


ThadHouse:
@SamCarlberg In order to use the new artifacts, I have to update CameraServer to use the new forceLoad functions, which throw an IOException. OnLoad for plugins doesn't throw IOException, so I have to try catch it in the method. Did I do it right? We would want to do the same forceLoad setup for nt as well, but I don't think we should do that in the same PR.

build.gradle.kts
```diff
@@ -17,30 +18,68 @@ buildscript {
...
+                        setProperty("repoKey", "release")
+                    } else {
+                        setProperty("repoKey", "development")
+                    }
+                    setProperty("username", System.getenv()["ARTIFACTORY_PUBLISH_USERNAME"])
+                    setProperty("password", System.getenv()["ARTIFACTORY_PUBLISH_PASSWORD"])
+                    setProperty("maven", true)
+                })
+            })
+            clientConfig.info.setBuildName("ShuffleBoard")
```
AustinShalit:
```suggestion
            clientConfig.info.setBuildName("Shuffleboard")
```

PR #636, opened 2019-11-24
Fix #615.

### LINKED ISSUE #615: ShadowJars published to maven always include linux binaries in addition to platform binaries
I noticed when trying to test Win32 shadowjar generation with a 64 bit jdk that the 64 bit binaries were being included in the final Jar. Looking at the maven artifacts, I noticed that the linux binaries are included in the Windows and mac jars. I think the reason why is the shadowjar plugin is including the build platform binaries because theyre part of the `compile` configuration, but also included the specified output platforms. I tried looking at a way to fix this, but the kotlin DSL kept getting in the way, and I wasn't able to figure out a solution.

carbotaniuman:
The fix for this does require users use 'runShadow' instead of 'run' to debug their applications in Gradle.

### DIFF
--- a/api-test-util/api-test-util.gradle.kts
+++ b/api-test-util/api-test-util.gradle.kts
@@ -5,5 +5,5 @@ plugins {
 }
 
 dependencies {
-    compile(project(":api"))
+    nativeProject(":api")
 }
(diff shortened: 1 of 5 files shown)


ThadHouse:
This still needs to support running from intellij as well ( I don't know how to do so myself, but people will whine otherwise.)

carbotaniuman:
Is that test failure linked to native issues? I don't see any LinkErrors being mentioned and IntelliJ won't tell me what test actually failed.

ThadHouse:
It likely is. Tests need access to the native libraries as well

Issue #649, opened 2020-02-09
Numeric value displayed as zero with a French environment


etiennebeaulac:
**Describe the bug**
When displaying a numeric value with a text view within a French environment (or any locale that uses the comma as the decimal separator instead of a period), the text view displays zero.

**To Reproduce**  
1. Start Shuffleboard with a French locale : `java -Duser.country=CA -Duser.language=fr ShuffleBoard.jar`
2. Connect Shuffleboard to a NetworkTable that contains a numeric value different from zero.
3. Drag this value onto the grid.

**Expected behavior**  
The text view should display the correct value and update accordingly.

**Screenshots**  
As you can see, the NetworkTables contains the good value (132.0) but ShuffleBoard displays it incorrectly.

![image](https://user-images.githubusercontent.com/12667152/74098508-eef21f00-4ae6-11ea-97e1-0326eac39c92.png)

**Desktop**
 - OS: Windows 10 10.0 amd64 (French)
 - Java version : 11.0.4+11 AdoptOpenJDK
 - Version : 2020.2.2

PR #650, opened 2020-02-09
Fix Numeric value displayed as zero with a French environment #649

etiennebeaulac:
# Overview
Fixes the bug reported in #649. I used DecimalFormat to force the use of the English locale. The maximum fraction digits of 340 was taken from the [DecimalFormat Javadocs](https://docs.oracle.com/javase/7/docs/api/java/text/DecimalFormat.html#setMaximumFractionDigits(int)), which is the maximum allowed for double values. This allows to show only the number of digits required, instead of the default 6 that were show by `String.format()`.

### DIFF
--- a/api/src/main/java/edu/wpi/first/shuffleboard/api/components/NumberField.java
+++ b/api/src/main/java/edu/wpi/first/shuffleboard/api/components/NumberField.java
@@ -1,6 +1,9 @@
 package edu.wpi.first.shuffleboard.api.components;
 
 import java.util.regex.Pattern;
+import java.util.Locale;
+import java.text.DecimalFormat;
+import java.text.DecimalFormatSymbols;
 
 /**
  * A type of text field that only accepts valid floating-point decimal numbers.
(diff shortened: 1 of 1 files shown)


sciencewhiz:
340 digits seems excessive

etiennebeaulac:
@sciencewhiz It does not really matter. Only significant digits are displayed : trailing zeros are removed. We could replace it by another arbitrary large number, but I think it is just best to put the real maximum (340), which does not limit the user, should the need ever arise.

AustinShalit:
Those zeros do get appended to the string when it is created costing some amount of time. The question is how much and does it matter.

carbotaniuman:
Can you change remove the `setMaximumFractionDigits()`? 3 is probably enough (default) as Shuffleboard can't really display more.

etiennebeaulac:
Shuffleboard can definitely display more than 3 decimal digits. If we use `setMaximumFractionDigits(3)`, numbers like `1.2e-4` will be displayed as zero.

> 
> 
> Those zeros do get appended to the string when it is created costing some amount of time. The question is how much and does it matter.

To address this concern, I ran a quick benchmark (multiple times) and the `DecimalFormat` (with 340 digits) ran faster than the old method (`String.format`), every time.

PR #653, opened 2020-03-03
Number slider with text

kevinclark:
# Overview
This PR allows the NumberSlider widget to display the value being specified by the slider. My team regularly uses this widget to tweak values in test mode and it's awkward to need to add an additional widget to display the value they're modifying. This by default shows the value and allows the display to be turned off via edit. Happy to swap the default if desired.

# Screenshots

Screen Shot 2020-03-02 at 5 01 04 PM


kevinclark:
Note that I pulled the padding and layout setup direct from NumberBar. Happy to tweak it, but figured uniformity was the right first guess.

SamCarlberg:
LGTM. Thanks for including screenshots!

kevinclark:
Thank you!

PR #656, opened 2020-03-22
Fix memory leak and graph being stuck in Shuffleboard's GraphWidget

carbotaniuman:
# Overview

This update fixes a memory leak in GraphWidget and lets the graph continue moving when no updates are made.

# Screenshots

https://youtu.be/_rQuhIWhAEs


plugins/base/src/main/java/edu/wpi/first/shuffleboard/plugin/base/widget/GraphWidget.java
```diff
@@ -123,7 +125,20 @@
     ThreadUtils.newDaemonScheduledExecutorService()
         .scheduleAtFixedRate(() -> {
           synchronized (graphWidgets) {
-            graphWidgets.forEach(GraphWidget::update);
+            graphWidgets.forEach(graphWidget -> {
```
SamCarlberg: I would put this in the update() function, or a helper method called by it

plugins/base/src/main/java/edu/wpi/first/shuffleboard/plugin/base/widget/GraphWidget.java
```diff
@@ -91,7 +93,7 @@
     updateFromArraySource(source);
   };
 
-  private final Map<Series<Number, Number>, SimpleData> realData = new HashMap<>();
+  private final Set<Series<Number, Number>> allSeries = new HashSet<>();
```
SamCarlberg: Not needed, just grab the Series objects from the JavaFX chart directly

SamCarlberg:
Keep in mind that this change loses all historical data after it leaves the visible graph; increasing the visible time will no longer make old data visible again

SamCarlberg:
As an aside, JavaFX charts really suck for realtime use. They use a ton of CPU, hence the low fixed update rate for graph widgets. I'd love to see a complete reimplentation with a more efficient chart library like https://github.com/GSI-CS-CO/chart-fx

carbotaniuman:
@SamCarlberg Understood for the first point. I'll probably get to a reimplementation of this widget once I fix a few other pending bugs/optimizations.

SamCarlberg:
The point I was trying to make is that some bugfixes or optimizations may not be needed when the underlying implementation changes, so it may be better to hold off on working on fixes for problems that soon may not even exist

SamCarlberg:
@carbotaniuman Yes, I am suggesting that the graph widget gets rewritten. Work like this PR may still be useful with a different chart library, but other bug fixes or optimizations might not be needed at all. It's a better use of time and effort to fix things that you know are still problems.

For example, a different chart implementation may have more efficient data storage, so making memory optimizations for the JavaFX chart data would be a waste of time

carbotaniuman:
@SamCarlberg 

```
  private final DoubleProperty yAxisMaxBound = new SimpleDoubleProperty(1);
  private final DoubleProperty visibleTime = new SimpleDoubleProperty(this, "Visible time", 30);
```

Any reason the second property has a lot more stuff than the first one?

SamCarlberg:
The second one was displayed in the properties dialog before I added the settings framework. It doesn't need those extra parameters any more

PR #657, opened 2020-03-24
Change GraphWidget to ChartFX

carbotaniuman:
# Overview

This widget changes the underlying graphing engine to chart-fx. This alleviates the memory cost of storing a large amount of doubles (as they are stored in unboxed form), as well as improving the refresh rate slightly.

We've also added an autoscrolling toggle to allow manual scrolling. Zooming and panning is allowed thanks to chart-fx.

# Video

https://youtu.be/8uW_hd7A7zI


plugins/base/src/main/java/edu/wpi/first/shuffleboard/plugin/base/widget/GraphWidget.java
```diff
@@ -1,5 +1,12 @@
 package edu.wpi.first.shuffleboard.plugin.base.widget;
 
+import com.google.common.collect.ImmutableList;
```
SamCarlberg: Import order should be

`edu.wpi.first.shuffleboard.*`
`edu.wpi.*`
[misc imports eg `com.*`, `de.*`, `edu.*`, grouped and alphabetical]
`java.*`
`javafx.*`

plugins/base/src/main/java/edu/wpi/first/shuffleboard/plugin/base/widget/GraphWidget.java
```diff
@@ -117,29 +111,35 @@
...
+  private static final long UPDATE_PERIOD = 200;
 
   static {
     ThreadUtils.newDaemonScheduledExecutorService()
         .scheduleAtFixedRate(() -> {
           synchronized (graphWidgets) {
+            // This method actually updates the graph (as seen by the user).
             graphWidgets.forEach(GraphWidget::update);
           }
         }, 500, UPDATE_PERIOD, TimeUnit.MILLISECONDS);
```
SamCarlberg: This can be changed. The default NT update rate is 10Hz - can we match that? Can we get to 30Hz?

If it can be made faster, look into making it a setting on the base plugin so users can get faster or slower refresh times

plugins/base/src/main/java/edu/wpi/first/shuffleboard/plugin/base/widget/GraphWidget.java
```diff
@@ -117,29 +111,35 @@
   /**
    * How often graphs should be redrawn, in milliseconds.
    */
-  private static final long UPDATE_PERIOD = 250;
+  private static final long UPDATE_PERIOD = 200;
```
SamCarlberg: The default NT update rate is 10Hz - can we match that? Can we get to 30Hz?

If it can be made faster, look into making it a setting on the base plugin so users can get faster or slower refresh times

plugins/base/src/main/java/edu/wpi/first/shuffleboard/plugin/base/widget/GraphWidget.java
```diff
@@ -282,111 +247,110 @@ private void updateFromArraySource(DataSource<double[]> source) {
...
     return series;
   }
 
-  private void updateBounds(long elapsed) {
-    xAxis.setUpperBound(elapsed);
-    removeInvisibleData();
-  }
-
   private void update() {
     FxUtils.runOnFxThread(() -> {
```
SamCarlberg: This is way too much code to live in a single lambda. Split it up into discrete functions

plugins/base/src/main/java/edu/wpi/first/shuffleboard/plugin/base/widget/GraphWidget.java
```diff
@@ -282,111 +247,110 @@ private void updateFromArraySource(DataSource<double[]> source) {
...
+          xAxis.maxProperty().set(dataSetMax.getAsDouble());
+          xAxis.minProperty().bind(xAxis.maxProperty().subtract(visibleTime.multiply(1e3)));
+          doubleDataSet.fireInvalidated(null);
+        } else {
+          xAxis.maxProperty().unbind();
+          xAxis.minProperty().unbind();
         }
       }
+
+      if (autoScrollToggle.isSelected() && globalMax.isPresent()) {
```
SamCarlberg: Consider putting `autoScrollToggle.isSelected()` into a helper method:

```java
private boolean isAutoScroll() {
  return autoScrollToggle.isSelected();
}
```

Alternatively, if autoscroll is made into a context menu `Action`, you'd need this method anyway to read from the boolean property of the setting.

plugins/base/src/main/java/edu/wpi/first/shuffleboard/plugin/base/widget/GraphWidget.java
```diff
@@ -62,24 +59,23 @@
...
-  private XYChart<Number, Number> chart;
+  private XYChart chart;
+  @FXML
+  private DefaultNumericAxis xAxis;
   @FXML
-  private NumberAxis xAxis;
+  private DefaultNumericAxis yAxis;
   @FXML
-  private NumberAxis yAxis;
+  private ToggleButton autoScrollToggle;
```
SamCarlberg: This looks out of place on the widget. Consider making this an `Action` in the context menu instead, or a `Setting` in the widget properties

plugins/base/src/main/java/edu/wpi/first/shuffleboard/plugin/base/widget/GraphWidget.java
```diff
@@ -191,54 +191,21 @@ public Number fromString(String string) {
...
-            synchronized (queueLock) {
-              chart.getData().forEach(s -> s.getData().clear());
-              queuedData.forEach((s, q) -> q.clear());
-              realData.forEach((s, d) -> d.clear());
-            }
-          });
-    });
+    ActionList.registerSupplier(root, () -> ActionList.withName(getTitle())
+        .addAction("Clear",
+            () -> chart.getDatasets().forEach(s -> {
```
SamCarlberg: Weird formatting. If you pull the "clear" functionality into its own method, this could look something like:


```java
ActionList.registerSupplier(
  root,
  () -> ActionList.withName(getTitle()).withAction("Clear", this::clear)
);
```

SamCarlberg:
### UI issues

Why do the chart labels have `[a.u.]` at the end? It's probably the default unit used by ChartFX, but it should be specified. X axis should always be `s` or `seconds`; Y axis might need to be user-configurable

The chart and scroll toggle button overflow the widget bounds at the default minimum size. Making the autoscroll option a context menu thing or a widget setting (or both) would be helpful in reducing clutter. Making it a setting would also mean it could be specified from the WPILib API.

image

The chart label text is always black, making it hard or impossible to read on dark themes
image

The chart stops growing vertically after a certain point, leaving useless whitespace at the bottom of the widget
image

carbotaniuman:
Is there any documentation on what specifically dark theme changes/where the code is located?

AustinShalit:
Nice getting dark mode working. We need to use stable versions of the software (no snapshots). Do you think you can convince that developer to push a release?

carbotaniuman:
@AustinShalit It looks like they're already moving to the next major version so I don't think that's going to happen, but I still have the rest of @SamCarlberg 's future review to address.

SamCarlberg:
### The Good

Performance seems much better now. I can run at 60fps fairly smoothly, but memory usage goes up really fast as a result. The GC seems to be able to handle it (goes from ~4GB heap to ~700MB, so there *is* a lot of garbage). Updates are much slower at larger heap sizes. The 10Hz configuration you've set should be OK, but I would like to see a global setting in the base plugin preferences to allow users to increase the update rate if they want.

### The Bad

- Axis labels are black on the midnight theme and aren't readable

- Unselecting Y-axis autoranging raises an error because you're binding the tick unit property, and the axis tries to set its own value to that bound property. Not sure if this is a bug in your code or in chartfx.

- Toggling visible datasets in the graph widget preferences has no effect

- Changing a small existing number widget (for example, a number field) to a graph makes the graph area overflow the bounds of the widget tile. This is probably more of an issue with how shuffleboard handles tiles and isn't something that needs to be addressed in this PR. (Though the simplest solution would probably be to check that the minimum size of the target type is smaller than the tile, and disable the option if the tile is too small)


Overall, I'm very happy with the changes you've made. Fix the small things and we can merge it

plugins/base/src/main/java/edu/wpi/first/shuffleboard/plugin/base/BasePlugin.java
```diff
@@ -202,4 +243,15 @@
...
   }
 
+  @Override
+  public List<Group> getSettings() {
+    return ImmutableList.of(
+            Group.of("Graph settings",
+                    Setting.of("Graph Update Rate",
+                            "How many hz (times a second) graph widgets update at. 25hz is considered the maximum tested",
+                            graphUpdateRate
+                    )
```
SamCarlberg, line 253: Be clear and avoid parentheticals in setting summaries:

```suggestion
                    Setting.of(
                      "Graph Update Rate",
                      "How many times a second graph widgets update at. Faster update rates may cause performance issues",
                      graphUpdateRate
                    )
```

plugins/base/src/main/java/edu/wpi/first/shuffleboard/plugin/base/BasePlugin.java
```diff
@@ -65,21 +70,57 @@
 import com.google.common.collect.ImmutableList;
 import com.google.common.collect.ImmutableMap;
 import com.google.common.collect.ImmutableSet;
+import javafx.beans.InvalidationListener;
+import javafx.beans.property.*;
```
SamCarlberg, line 74: I think this line is breaking checkstyle

plugins/base/src/main/java/edu/wpi/first/shuffleboard/plugin/base/BasePlugin.java
```diff
@@ -65,21 +70,57 @@
...
 )
 @SuppressWarnings("PMD.CouplingBetweenObjects")
 public class BasePlugin extends Plugin {
+  private final Preferences preferences = Preferences.userNodeForPackage(getClass());
+
+  private final IntegerProperty graphUpdateRate = new SimpleIntegerProperty(this, "graphUpdateRate", 10);
+  private final InvalidationListener graphSaver = __ -> PreferencesUtils.save(graphUpdateRate, preferences);
+
+  // only written to one thread at a time
+  private volatile ScheduledFuture<?> currentFuture;
```
SamCarlberg, line 98: Maybe track this somewhere else? It doesn't make sense for the plugin to have to manage custom logic for a widget's configuration.

carbotaniuman: The way it works currently every graph updates at the same rate. It's a holdover from the previous impl.

SamCarlberg: I'm talking more from a code organization standpoint. Its better to encapsulate the graph widget updater logic in its own class instead of managing it in the base plugin

PR #662, opened 2020-06-19
Change PID to support new PIDController

carbotaniuman:
# Overview

This removes the `enabled` toggle in PIDControlelrWidget and makes it compatible with the new PIDController.

# Screenshots


sciencewhiz:
The would make enable not work on the old PIDController, whereas currently it's just not used by new PIDController. I don't think it's worth breaking old PIDController, when the new PIDController works fine now.

carbotaniuman:
Would we not want a new PIDCommand widget then? Seems like an oversight to not be compatible with new commands.

Daltz333:
The new PIDController works fine with the old widget.

PR #663, opened 2020-07-02
Updates to new wpilib native libraries extraction method and switches back to groovy dsl

ThadHouse:
We're switching to a new method of extracting wpilib native libraries that is more reliable and more universally configured across all of our tools. There is now a plugin to handle all of this configuration across all tools.

Part of this change removed the possibility of generating alternate platform jars on a single system. This was broken anyway before, and removing it massively cleans up the build

This PR also reverts away from using the Kotlin DSL. It has many issues, mostly around compatibility with VS Code and none of us current WPILib developers know it, whereas we know groovy, since everything uses it.


app/src/main/java/edu/wpi/first/shuffleboard/app/Main.java
```diff
@@ -16,7 +20,10 @@
   private static final Logger logger = Logger.getLogger(Main.class.getName());
 
   @SuppressWarnings("JavadocMethod")
-  public static void main(String[] args) {
+  public static void main(String[] args) throws IOException {
+    WPIUtilJNI.Helper.setExtractOnStaticLoad(false);
+    CombinedRuntimeLoader.loadLibraries(Main.class, "wpiutiljni");
```
SamCarlberg, line 25: Why does this have to be called in Main?

plugins/base/src/main/java/edu/wpi/first/shuffleboard/plugin/base/data/fms/FmsInfo.java
```diff
@@ -69,7 +69,7 @@ public FmsInfo(Map<String, Object> map) {
         Maps.<String, Number>get(map, MATCH_NUMBER).intValue(),
         Maps.<String, Number>get(map, REPLAY_NUMBER).intValue(),
         MatchType.fromOrdinal(Maps.<String, Number>get(map, MATCH_TYPE).intValue()),
-        Maps.get(map, IS_RED_ALLIANCE) ? Alliance.RED : Alliance.BLUE,
+        (Boolean)Maps.get(map, IS_RED_ALLIANCE) ? Alliance.RED : Alliance.BLUE,
```
SamCarlberg, line 72: This change seems irrelevant to this PR

ThadHouse: It failed to compile with if I didn't do this. Not sure why.

app/src/main/java/edu/wpi/first/shuffleboard/app/Main.java
ThadHouse: It's the base libs, so it either needs to be called in main or at the very beginning when javafx initialized. Main seemed like the better option since it's the actual entry point.

ThadHouse: Part of this change is that automatic extraction will no longer happen anymore. This portion loads wpiutil, which then allows the plugins to load their own libraries

plugins/base/src/main/java/edu/wpi/first/shuffleboard/plugin/base/data/fms/FmsInfo.java
ThadHouse: Ok, must have just been a weird bug. Reverted.

ThadHouse:
I merged in master, so this is ready to go for final review and then merge.

PR #668, opened 2020-09-13
Fix shuffleboard not closing

### LINKED ISSUE #666: App does not close correctly
Ever since the Graph update, shuffleboard does not close correctly. This needs to be fixed before we release.

carbotaniuman:
Fixes #666.

### DIFF
--- a/app/src/main/java/edu/wpi/first/shuffleboard/app/Shuffleboard.java
+++ b/app/src/main/java/edu/wpi/first/shuffleboard/app/Shuffleboard.java
@@ -139,16 +139,21 @@ public class Shuffleboard extends Application {
       alert.getButtonTypes().setAll(ButtonType.YES, ButtonType.NO, ButtonType.CANCEL);
       alert.getDialogPane().setHeaderText("Save the current layout before closing?");
       alert.showAndWait().ifPresent(bt -> {
+        if (bt == ButtonType.CANCEL) {
+          // cancel the close request by consuming the event
+          closeEvent.consume();
+          return;
+        }
+
         if (bt == ButtonType.YES) {
           try {
             mainWindowController.save();
           } catch (IOException ex) {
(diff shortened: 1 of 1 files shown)


Daltz333 (approved):

app/src/main/java/edu/wpi/first/shuffleboard/app/Shuffleboard.java
```diff
@@ -139,16 +139,21 @@ public void start(Stage primaryStage) {
...
           } catch (IOException ex) {
             logger.log(Level.WARNING, "Could not save the layout", ex);
           }
-        } else if (bt == ButtonType.CANCEL) {
-          // cancel the close request by consuming the event
-          closeEvent.consume();
         }
+
+        Platform.exit();
         // Don't need to check for NO because it just lets the window close normally
```
sciencewhiz, line 157: should the comment go above exit command?

carbotaniuman: Fixed.

AustinShalit:
So Platform should auto close when all stages & threads close. This is more of a patch than the real fix. Do you have any ideas what thread the graphing API still has running that is not being closed?

carbotaniuman:
I checked out the thread list after this, and every thread left was either a JDK or a daemon thread, which shouldn't impede shutdowns.

carbotaniuman:
With chart-fx:
image

Without:
image

It's a real ???? right now.

carbotaniuman:
@AustinShalit debugged it - https://github.com/GSI-CS-CO/chart-fx/blob/8485ed6cc44f2eb531cfe55c10411c0a1da35c6e/chartfx-chart/src/main/java/de/gsi/chart/utils/FXUtils.java#L39

AustinShalit (approved):

PR #671, opened 2020-09-13
Use DecimalFormat to format numbers

### LINKED ISSUE #649: Numeric value displayed as zero with a French environment

carbotaniuman:
Fixes #649.

### DIFF
--- a/api/src/main/java/edu/wpi/first/shuffleboard/api/components/NumberField.java
+++ b/api/src/main/java/edu/wpi/first/shuffleboard/api/components/NumberField.java
@@ -1,6 +1,9 @@
 package edu.wpi.first.shuffleboard.api.components;
 
 import java.util.regex.Pattern;
+import java.util.Locale;
+import java.text.DecimalFormat;
+import java.text.DecimalFormatSymbols;
 
 /**
  * A type of text field that only accepts valid floating-point decimal numbers.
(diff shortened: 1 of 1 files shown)


AustinShalit (approved):

etiennebeaulac:
@carbotaniuman @AustinShalit I tried the code on my machine, and I think that the default number of decimal digits is zero. Therefore, no more decimals are currently displayed, which seems problematic.

Also, as I said in #650, even if the default is 3, this means that all decimals after the third will be removed and numbers smaller than 0.001 will be rounded to zero. That does not seems to right me either.

PR #701, opened 2021-07-27
Show exit save prompt only if not saved

noamzaks:
# Overview
Show the `Save the current layout before closing?` prompt only if the current layout is actually not saved.


carbotaniuman (approved):
Thanks for your changes!

AustinShalit (approved):

PR #703, opened 2021-08-11
Add Field2d Widget

### LINKED ISSUE #700: Add Field2d to Shuffleboard
**Is your feature request related to a problem? Please describe.**  
No

**Describe the solution you'd like**  
As part of the PC Simulation Framework, there was a new object created called Field2d.  Since all the data for the table is on NetworkTables, it should be "easy" to port over to Shuffleboard.  

**Describe alternatives you've considered**  
Field2d works great in SIL, it would be nice for the competition use.

**Additional context**  
See https://docs.wpilib.org/en/stable/docs/software/wpilib-tools/glass/field2d-widget.html

noamzaks:
# Overview
This PR aims to visualize Field2d (in a way similar to Glass).
Closes #700.

# Status
- I think the widget should use some parts of PathWeaver regarding loading game backgrounds, as well as use PathWeaver's resources for the up-to-date fields. I think there a few ways to go about this:
  - Add PathWeaver as a runtime dependency for the base plugin
  - Separate Field2d into its own plugin, and add PathWeaver as a dependency there
  - Separate some parts of PathWeaver into a shared artifact (I'm guessing this is a big maintainability issue tho)
- I have not succeeded yet in adding a ChoiceBox to the settings pane of the widget (for choosing the game)


plugins/base/src/main/java/edu/wpi/first/shuffleboard/plugin/base/data/FieldData.java
```diff
@@ -0,0 +1,90 @@
...
+      }
+
+      double[] doubles;
+      if (map.get(key) instanceof byte[]) {
+        byte[] data = (byte[]) map.get(key);
+        doubles = new double[data.length / Double.BYTES];
+        for (int i = 0; i < doubles.length; i++) {
+          ByteBuffer buffer = ByteBuffer.allocate(Double.BYTES);
+          // this implementation assumes Double.BYTES is 8
+          buffer.put(new byte[]{data[8*i], data[8*i+1], data[8*i+2], data[8*i+3], data[8*i+4], data[8*i+5], data[8*i+6], data[8*i+7]});
```
carbotaniuman, line 65: Probably better to use the 3-arg overload of `buffer.put` here.

plugins/base/src/main/java/edu/wpi/first/shuffleboard/plugin/base/widget/FieldWidget.java
```diff
@@ -0,0 +1,273 @@
...
+      return this.name().substring(1).replaceAll("_", " ");
+    }
+  }
+
+  private final Property<Game> game =
+          new SimpleObjectProperty<>(Game.A2021_Infinite_Recharge);
+
+  private final Map<String, Paint> colors = new HashMap<>();
+
+  private static final double ROBOT_SIZE = 30;
```
carbotaniuman, line 85: Can we make this configurable?

plugins/base/src/main/java/edu/wpi/first/shuffleboard/plugin/base/data/FieldData.java
```diff
@@ -0,0 +1,101 @@
...
+      }
+
+      double[] doubles;
+      if (entry.getValue() instanceof byte[]) {
+        byte[] data = (byte[]) entry.getValue();
+        doubles = new double[data.length / Double.BYTES];
+        for (int i = 0; i < doubles.length; i++) {
+          ByteBuffer buffer = ByteBuffer.allocate(Double.BYTES);
+          // this implementation assumes Double.BYTES is 8
+          buffer.put(new byte[]{
```
carbotaniuman, line 69: You can use the 3-arg overload of ByteBuffer#put here. Alternatively, ByteBuffer#wrap would probably be the right move.

plugins/base/src/main/java/edu/wpi/first/shuffleboard/plugin/base/widget/FieldWidget.java
```diff
@@ -0,0 +1,273 @@
...
+  private void initialize() {
+    robot.setImage(
+            new Image(getClass().getResource("field/robot.png").toExternalForm()));
+    robot.setFitWidth(ROBOT_SIZE);
+    robot.setFitHeight(ROBOT_SIZE);
+    setGame(game.getValue());
+    game.addListener(__ -> {
+      setGame(game.getValue());
+      centerImage();
+      updateRobotPosition();
```
carbotaniuman, line 97: Is the coordinate system based on the image and root height/width? Alternatively, does this need to be called here?

plugins/base/src/main/java/edu/wpi/first/shuffleboard/plugin/base/widget/FieldWidget.java
```diff
@@ -0,0 +1,273 @@
...
+  private void updateRobotPosition() {
+    robot.setTranslateX(
+            transformX(dataOrDefault.get().getRobot().getX(), ROBOT_SIZE / 2));
+    robot.setTranslateY(
+            transformY(dataOrDefault.get().getRobot().getY(), ROBOT_SIZE / 2));
+    robot.setRotate(-dataOrDefault.get().getRobot().getDegrees());
+  }
+
+  private void updateObjects() {
+    pane.getChildren().removeIf(n -> n instanceof Circle);
```
carbotaniuman, line 231: Are there any perf issues throwing out all the circles and readding them?

plugins/base/src/main/java/edu/wpi/first/shuffleboard/plugin/base/widget/FieldWidget.java
noamzaks: The circle placements are dependent on the field width and height as well as the image fitting and field startX, endX, startY and endY, so if this wasn't called it wouldn't be placed correctly

plugins/base/src/main/java/edu/wpi/first/shuffleboard/plugin/base/data/FieldData.java
noamzaks: Thanks! That does seem more elegant 😀

noamzaks:
I would like to point out here that @PeterJohnson mentioned that when NT4 is implemented this will need to be slightly refactored, by getting the objects through an array (as entries will no longer be deleted)

PR #742, opened 2022-11-23
Update Power Distribution Widget to support 24 channels

sciencewhiz:


PeterJohnson (approved):

PeterJohnson:
Ideally this would show different displays for the different number of channels (to match the physical device), but I know that's substantially more work.

sciencewhiz:
Especially for shuffleboard it would be hard. SD would be easier. However when I was doing it in simulation it looked like 24 channels of data was passed regardless, so not sure that it's possible.

PeterJohnson:
Interesting.  So we'd need to do something like actually pass the number of channels through NT so the dashboard even knows how many there are.  That's likely worth doing even if we don't support it in Shuffleboard, as it could be used in other dashboards.

PR #759, opened 2023-02-14
Check for invalid NT value

### LINKED ISSUE #758: NT NPE
No errors appear in the UI, but these errors are printed a lot when running via CLI.
```
WARNING: Uncaught exception on JavaFX Application Thread
java.lang.NullPointerException: Cannot invoke "Object.getClass()" because the return value of "edu.wpi.first.networktables.NetworkTableValue.getValue()" is null
        at edu.wpi.first.shuffleboard.plugin.networktables.util.NetworkTableUtils.dataTypeForEntry(NetworkTableUtils.java:55)
        at edu.wpi.first.shuffleboard.plugin.networktables.sources.NetworkTableSourceType.dataTypeForSource(NetworkTableSourceType.java:154)
```

@PeterJohnson is something here wrong with how NT4 is used?

Starlight220:
Fixes #758

### DIFF
--- a/plugins/networktables/src/main/java/edu/wpi/first/shuffleboard/plugin/networktables/util/NetworkTableUtils.java
+++ b/plugins/networktables/src/main/java/edu/wpi/first/shuffleboard/plugin/networktables/util/NetworkTableUtils.java
@@ -55,6 +55,8 @@ public final class NetworkTableUtils {
       var ntValue = rootTable.getEntry(normalKey).getValue();
       if (ntValue.isValid()) {
         return DataTypes.getDefault().forJavaType(ntValue.getValue().getClass()).get();
+      } else {
+        return null;
       }
     }
     if (rootTable.containsSubTable(normalKey)) {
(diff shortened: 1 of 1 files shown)


plugins/networktables/src/main/java/edu/wpi/first/shuffleboard/plugin/networktables/util/NetworkTableUtils.java
```diff
@@ -52,7 +52,10 @@ public static DataType dataTypeForEntry(String key) {
       return DataTypes.Map;
     }
     if (rootTable.containsKey(normalKey)) {
-      return DataTypes.getDefault().forJavaType(rootTable.getEntry(normalKey).getValue().getValue().getClass()).get();
+      var ntValue = rootTable.getEntry(normalKey).getValue();
+      if (ntValue.isValid()) {
+        return DataTypes.getDefault().forJavaType(ntValue.getValue().getClass()).get();
+      }
```
PeterJohnson, line 58: It feels like this shouldn’t fall through (to the if subtable comparison), and instead should immediately return null.

PR #760, opened 2023-02-14
Use NT4 metadata

Starlight220:
Still far from ready.


Starlight220:
This is becoming rather large and disorganized, so I'm breaking it out into multiple PRs.

PR #761, opened 2023-02-15
Remove 25th channel from PDP widget

Starlight220:


SamCarlberg:
You'll need to remove it from the FXML file as well

Starlight220:
It's not in the FXML, that's the point.

SamCarlberg (approved):

PR #767, opened 2023-02-20
Create tabs for root tables

Starlight220:
Create tabs based on root tables, not only those inside `/Shuffleboard/`. This is a prerequisite for #760, and is already rebased on #766.
Also removed the default LiveWindow and SmartDashboard tables; these will be created from their NT tables if those exist. (Though an empty LW tab appears to always be created due to the always-present `.status/` "hidden" subtable.)


Daltz333:
I disagree with creating tabs based on root tables. I think forcing tabs for NT devices such as PhotonVision, LL, and other devices will lead to clutter and annoyance as a user.

Starlight220:
You raise a good point that I haven't thought of.

PR #771, opened 2023-03-12
Don't switch to selected tab on value change

### LINKED ISSUE #745: Shuffleboard continuously jumps back to start tab when using selectTab and kGyro
**Describe the bug**  
When using WPILIB code to create the Shuffleboard layout and select the initial tab, if one of the tabs contains a BuiltinWidget.kGyro, then you will no be able to change tabs, because the tab will switch back to the start tab right away.
It does not seem to matter which tab the Gyro is on (start tab or not).

**To Reproduce**  
I created a very simple Robot.java to duplicate this problem.
Compile and deploy the attached Robot.java (you will also need the https://dev.studica.com/releases/2023/NavX.json vendordeps).

You may or may not need to restart shuffleboard after deploying (depending on your previous layout).
It will automatically start on Tab Two as expected.  But if you try to select Tab one, it will bounce back to Tab Two automatically.

Removing the kGyro widget eliminates the problem, so there is some sort of interaction with selectTab and kGyro.

 - OS: Windows 10 10.0 amd64
 - Java version: 17.0.5+8 -- Eclipse Adoptium
 - Version: 2023.1.1 built on 1/2/23

**Additional context**  
[Robot_java.txt](https://github.com/wpilibsuite/shuffleboard/files/10361374/Robot_java.txt)

Starlight220:
Fixes #745

For some reason, the tab structure listeners were called also on value changes (and not just metadata changes). I've done some testing to ensure this doesn't break functionality, but I may have missed something.

### DIFF
--- a/plugins/networktables/src/main/java/edu/wpi/first/shuffleboard/plugin/networktables/NetworkTablesPlugin.java
+++ b/plugins/networktables/src/main/java/edu/wpi/first/shuffleboard/plugin/networktables/NetworkTablesPlugin.java
@@ -42,7 +42,7 @@ import javafx.beans.value.ChangeListener;
 @Description(
     group = "edu.wpi.first.shuffleboard",
     name = "NetworkTables",
-    version = "2.3.1",
+    version = "2.3.2",
     summary = "Provides sources and widgets for NetworkTables"
 )
 public class NetworkTablesPlugin extends Plugin {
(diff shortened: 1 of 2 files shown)


Starlight220:
This has an issue with repeated calls from code to not work due to NT duplicate removal. It'll need the `KeepDuplicates` flag set on the robot-side publisher and possibly also the shuffleboard-side subscriber.

Starlight220:
I've extracted the tab selection to a separate listener (similar to how the tab array is separate), and now everything works correctly once the robot-code side changes in wpilibsuite/allwpilib#5198 are applied.

PeterJohnson (approved):

PR #774, opened 2023-03-19
Add DataSource.equals()

### LINKED ISSUE #720: Graphs get duplicate entries on code deploy
**Describe the bug**  
When graphing network table values after redeploying code, the graph will have two copies of the same NT value with the same name. I first discovered this when graphing two values on one graph, but the bug is present even when values are in separate graphs. 

In my code I publish two doubles with this code.

https://github.com/FRC-Sonic-Squirrels/swerve-template/blob/0c9cf83f6509f439c864c6501059f33c5e1afcee/src/main/java/frc/robot/commands/DriveWithSetRotationCommand.java#L119-L120

**To Reproduce**  
1. publish value to shufflboard (see link to example code above)
2. create entry on Shuffboard page and choose "disaplay as graph"
3. observe graph works
4. redeploy code from VS Code
5. now there are two numbers graphed with identical NT names
6. redeploy code
7. now the graph has 3 identical NT entries all graphed
8. repeat

**Expected behavior**  
graph should only have one copy of the Network Table value in the graph

**Screenshots**  
See screen shot
IMG_2281

 - OS: Windows
  - Roborio 1.0, Image FRC_roborio_2022_v3.0
  - DS 22.0
  - Lib Java 2022.1.1
  - Pidgeon 22.0
  - PDP 1.40
  - CANCoder 22.0.0.0
  - TalonFX 22.0.0.0

**Additional context**  
It feels like Shuffleboard is using more CPU and memory than prior to the 2022.1.1 release, I don't have a stats to back this up. Just the impression that on the same laptop with a far more complex Shuffleboard layout on the 2021 code base, Shuffleboard didn't cause my laptop to start "weezing" (i.e. run the fan nonstop and kill the battery).

Starlight220:
Should fix #720, and maybe even the disconnected source issues.

### DIFF
--- a/api/src/main/java/edu/wpi/first/shuffleboard/api/sources/AbstractDataSource.java
+++ b/api/src/main/java/edu/wpi/first/shuffleboard/api/sources/AbstractDataSource.java
@@ -6,6 +6,7 @@ import edu.wpi.first.shuffleboard.api.properties.AtomicBooleanProperty;
 import edu.wpi.first.shuffleboard.api.widget.Sourced;
 
 import java.util.Collections;
+import java.util.Objects;
 import java.util.Set;
 import java.util.WeakHashMap;
 
@@ -116,4 +117,8 @@ public abstract class AbstractDataSource<T> implements DataSource<T> {
     return !clients.isEmpty();
   }
 
+  @Override
+  public boolean equals(Object o) {
+    return o instanceof DataSource<?> && Objects.equals(((DataSource<?>) o).getId(), this.getId());
+  }
 }
--- a/api/src/main/java/edu/wpi/first/shuffleboard/api/sources/AbstractDataSource.java
+++ b/api/src/main/java/edu/wpi/first/shuffleboard/api/sources/AbstractDataSource.java
@@ -119,7 +119,7 @@ public abstract class AbstractDataSource<T> implements DataSource<T> {
 
   @Override
   public boolean equals(Object o) {
-    return o instanceof DataSource<?> && Objects.equals(((DataSource<?>) o).getId(), this.getId());
+    return o.getClass().equals(this.getClass()) && Objects.equals(((DataSource<?>) o).getId(), this.getId());
   }
 
   @Override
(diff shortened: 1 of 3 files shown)


api/src/main/java/edu/wpi/first/shuffleboard/api/sources/AbstractDataSource.java
```diff
@@ -116,4 +117,13 @@ public boolean hasClients() {
     return !clients.isEmpty();
   }
 
+  @Override
+  public boolean equals(Object o) {
+    return o instanceof DataSource<?> && Objects.equals(((DataSource<?>) o).getId(), this.getId());
```
SamCarlberg, line 122: Check that the classes are equal, not just instanceof the interface.  This might blow up on interactions with destroyed sources

Starlight220: How do we want this to interact with destroyed sources?

I think they don't inherit from AbstractDataSource, so they wouldn't have this impl anyway. Should they?

SamCarlberg: Ah, if they don't extend from this class then I'm fine with the PR as is

SamCarlberg (approved):

api/src/main/java/edu/wpi/first/shuffleboard/api/sources/AbstractDataSource.java
Starlight220: I haven't looked too much at the DestroyedSource impl, but theoretically this comparison might return true if the rhs was a destroyed source with the same URI. If that's possible, is it what we want?

Starlight220: Comparison with a destroyed source in rhs did indeed return true. Changed to check class, and added some tests.

Starlight220:
Can someone with the relevant perms rerun CI? The failures are due to a Bintray outage.

PR #789, opened 2023-12-03
Remove javadoc search workaround

sciencewhiz:
Fix was backported to Java 11


sciencewhiz:
builds will fail until #788 is merged

PR #790, opened 2023-12-03
Renable UI tests since TestFX was upgraded

sciencewhiz:


sciencewhiz:
builds will fail until #788 is merged

PeterJohnson (approved):

PR #805, opened 2024-06-12
Add persistent alerts widget

jwbonner:
Related to https://github.com/wpilibsuite/allwpilib/issues/6573 and https://github.com/wpilibsuite/allwpilib/pull/6735, based on the 6328 [NetworkAlerts](https://github.com/Mechanical-Advantage/NetworkAlerts) plugin.


plugins/base/src/main/java/edu/wpi/first/shuffleboard/plugin/base/data/types/AlertsType.java
```diff
@@ -0,0 +1,41 @@
...
+   */
+  public static final AlertsType Instance = new AlertsType();
+
+  private AlertsType() {
+    super(TYPE_NAME, AlertsData.class);
+  }
+
+  @Override
+  public Function<Map<String, Object>, AlertsData> fromMap() {
+    return map -> new AlertsData((String[]) map.getOrDefault("errors", new String[] {}),
```
SamCarlberg, line 32: You can use the `Maps` utility class to avoid casting

```java
Maps.getOrDefault(map, "errors", new String[0])
```

plugins/base/src/main/java/edu/wpi/first/shuffleboard/plugin/base/data/AlertsData.java
```diff
@@ -0,0 +1,93 @@
...
+    }
+    for (String text : warnings) {
+      collection.add(new AlertItem(AlertType.WARNING, text));
+    }
+    for (String text : infos) {
+      collection.add(new AlertItem(AlertType.INFO, text));
+    }
+    if (collection.size() == 0) {
+      collection.add(new AlertItem(AlertType.LOCAL, "(Nothing to report)"));
+    }
```
SamCarlberg, line 65: This would be better as a placeholder[1] on the widget instead of being baked into the data object. It would also make the `LOCAL` alert type unnecessary

[1] https://docs.oracle.com/javase/8/javafx/api/javafx/scene/control/ListView.html#placeholderProperty--

plugins/base/src/main/resources/edu/wpi/first/shuffleboard/plugin/base/widget/AlertsWidget.fxml
```diff
@@ -0,0 +1,14 @@
...
+<?import javafx.scene.control.Label?>
+<?import javafx.scene.control.ListView?>
+<?import javafx.scene.image.ImageView?>
+<?import javafx.scene.image.Image?>
+<?import javafx.scene.layout.*?>
+<VBox xmlns="http://javafx.com/javafx"
+      xmlns:fx="http://javafx.com/fxml"
+      fx:controller="edu.wpi.first.shuffleboard.plugin.base.widget.AlertsWidget"
+      fx:id="root">
+      <ListView fx:id="list"/>
```
SamCarlberg, line 13:
```suggestion
      <ListView fx:id="list">
          <placeholder>
              <Label text="(Nothing to report)"/>
          </placeholder>
      </ListView>
```

plugins/base/src/main/java/edu/wpi/first/shuffleboard/plugin/base/widget/AlertsWidget.java
```diff
@@ -0,0 +1,159 @@
...
+                break;
+              case WARNING:
+                imageView.setImage(warningIcon);
+                break;
+              case INFO:
+                imageView.setImage(infoIcon);
+                break;
+              default:
+                break;
+            }
```
SamCarlberg, line 73: Java 17 language features make this a lot nicer

```java
var icon = switch (item.type) {
  case ERROR -> errorIcon;
  case WARNING -> warningIcon;
  case INFO -> infoIcon;
}

imageView.setImage(icon);
```

SamCarlberg (changes requested):

plugins/base/src/main/java/edu/wpi/first/shuffleboard/plugin/base/data/AlertsData.java
jwbonner: Fixed, with a slight style tweak so that the background color doesn't change when the list disappears

plugins/base/src/main/java/edu/wpi/first/shuffleboard/plugin/base/widget/AlertsWidget.java
jwbonner: It looks like we would need to update to at least checkstyle 8.36 to support Java 14 switches, but there are compatibility issues with the current config files. It seems like that should be fixed separately, but is probably not blocking?

SamCarlberg: Yeah, that could be fixed separately

SamCarlberg (approved):

PR #807, opened 2024-06-20
Puncture the bloat balloon

SamCarlberg:
# Overview

Removes various sources of high CPU usage:

- The live network table preview has been removed due to the high frequency of node redraws. Spot checking data will now need to be done by dragging a data widget into a tab. Generic table data should still be displayable with the generic tree table widget. The NT source preview now displays the data type of a topic instead of its current value.
- The subscribe-to-all-NT-data behavior has been changed to only subscribe to topic information. This should cut down on bandwidth and some CPU usage
- The widget gallery has been removed with no replacement. Drawing the large number of dummy widgets took a surprising CPU load and occasionally caused problems at startup when third party widgets from plugins would crash

And some bug fixes:
- Fixes widgets getting permanently disabled after network disconnects
  - This was caused by the placeholder destroyed sources not always being removed when restored, causing the widget to think it still had a disconnected data source


Starlight220:
How coupled are these changes? I think it'd be better to split it into different PRs.

SamCarlberg:
> How coupled are these changes? I think it'd be better to split it into different PRs. 

The live NT preview and widget gallery changes are both resulting from profiling the JavaFX scene graph. The NT subscribe-to-all change naturally falls out of that, since there's no need to have subscriptions that are never used

TheTripleV:
Live NT preview is pretty popular. Should Shuffleboard get a pinned message at the top of the tree view pane saying to use AdvantageScope for this?

PeterJohnson (approved):

